PR #31, opened 2016-02-13
Pull Request to close issue #30

### LINKED ISSUE #30: Remove dependency on Git for building
Solution for #4 only works when building from a local git repository. This creates a dependency from Git, that prevents building from sources downloaded from a tar ball. 

Also, now CMake > 2.811 is required for getting build date/time.

It is necessary to review the solution for #4 and do it in a different way.

cecilios:
Closes #30: Remove dependency from Git for building


kolewu:
It's not a good idea to have a version header with a commit hash in the repo, because this means you have to update it with every commit and this is not possible cause you don't know the commit hash in advance.

The version (or at least the commit hash part) should be created from the build tool. When building from git, the former solution was sufficient, so I think the remaining question is: What to do with the commit hash variable when not building from repository.

With [semantic versioning](http://semver.org/) the answer is not really easy, when you want to retain the meaning (bugfix on PATCH, features on MINOR and incompatible API changes on MAJOR) of the version numbers while developing and commiting.

I would suggest to follow the semver spec and add the commit hash as build information separated by a `+` like in `0.17.7+sha1` and incrementing MINOR and PATCH only for a release (git tag). For releases the commit hash can be left out, because the commit is clearly identified with the version (without hash) that equals a git tag.

With this scheme builds from repository and builds from releases (tags) can be clearly separated and you don't have to maintain (and thus change) the version file for every single commit.

cecilios:
Thanks for the feedback. 

When moving to Git I had to change the workflow and the way in which I maintain version information. 

My aim has been always to follow the semantic version specification (http://semver.org/) but I didn't realize that build information is separated by a +. No problem to change this. I have opened issue #32. In any case:

- sha1 does not form part of the package version number. It is always major.minor.patch (i.e. liblomse_0.17.34_amd64.deb).
- sha1 only appears in the string returned by method LomseDoorway::get_version_long_string(). So this is the code to be changed.

As to your other point (not a good idea to have a version header with a commit hash in the repo) these are my thoughts, please take them as a brain storming, not my final position  ;-)

My objective is to have master repo always buildable and usable, without having to do official releases. Of course, from time to time I will do an official release, but it will be just a point in time in which I decide to build and release binary packages (i.e for Ubuntu PPA), thats all!. A release takes time and work (building packages, writing news, posting news, updating website, etc.) so my aim is to reduce management work by keeping master branch always ready to be usable: every merge is a release.

A second objective is to minimize the work required for maintaining version information, obtaining it from Git if possible, but at the same time, retaining all necessary information for identifying code base state when someone reports a bug.

With all these in mind, I will only care of maintaining MAJOR and MINOR numbers, that will be indicated by a tag. 

As to the patch number, I will not retain semver meaning in PATCH number. For semver PATCH number denotes **only** bug fixes but, what if a compatible change without adding more functionality of fixing a bug (i.e. changes in documentation, refactoring code without changing behavior, etc.)?

So for Lomse, a PATCH number change denote **any compatible change** from previous release, not necessarily a bug fix. And, as you can imaging, PATCH number is the number of commits since the last tag.

For adding version information my idea is to use tags and Git information ('git --describe' command). My first idea was what you suggest: adding the version information when building from git. I tried this approach but I found two serious problems: 

1. Builds created from source packages will not have version information (or not updated). Only builds created from repo clones will have updated version information.
2. Build system will have a dependency on Git package.

For solving this, I was forced to maintain version information on a header file. For automating this I take advantage of these facts:

1. Direct commits on master branch are FORBIDDEN (with very few exceptions when I have to do a very simple minor change that do not affect the code, i.e.: to add a document file). Therefore, it can be assumed that the master repo is always updated via a merge after a pull request.

2. As merge == new release, a merge is the ideal point for updating version information.

Therefore, my current workflow for doing a merge includes these steps:

1. Evaluate if there are API/ABI changes and, is so, add a new tag (MAJOR/MINOR).
2. Do the merge
3. update lomse_version.h (direct commit on master branch).

As you can see, SHA is only updated when a new release, not with every commit. And as you point out, it is not really needed because PATCH should point to it. But I prefer to keep SHA: it is free and could be of help.

Probably current solution can be improved and ideas are welcome. But I would like to automate, as much as possible, the work for maintaining version information, and ensuring that when a user reports a bug I can feasible determine the code base he/she is using.

Comments welcome!

kolewu:
Thanks for your detailed answer. I think we are in the same boat here!

I better continue this discussion in the merge request #33.

Issue #32, opened 2016-02-27
change version long string to follow semver specification


cecilios:
To follow the semantic versioning specification (http://semver.org/), add the commit hash as build information separated by a + like in 0.17.7+sha1

PR #33, opened 2016-02-28
get version from include and buildinfo from git

kolewu:
* better parsing of include file
* if build from git repo add sha1 and dirty flag

This code (like the current one) has one drawback: version information will only be updated when running cmake. So if you change the version header file `include/lomse_version.h` it will not be used by the next _normal_ build.

This change belongs to #31 and #32.


kolewu:
I only use the hash when building from a git checkout. And I consider the fact, that releases on github AFAIK are only exports of tags, so there is no possibility to make extra files for the release package that are not part of the repository.

Now the workflow can be exactly as you described it:

1. make some changes inside a branch
2. when ready merge and increment the patch version

Everytime when you feel to make a release, just tag the current version on master. And I would suggest to always use the full version (0.17.0 and not 0.17).

When you think, there are some good features together, then change the minor-version, add some documentation and tag it.

Maybe with this workflow you can easily get some roadmap planning without too much extra work. Just define what features or fixes should be part of the next minor version and if they are ready, then just tag this version right from the master.

With the included version you can see and distinguish clearly:

- when only the version is given as long version, it is a build from a release package
- when there is a build information appended it's build from a git checkout with the given hash
- additionally you can see if the source has been changed before building

Unfortunately there is one flaw with this implementation: the version information is only changed when cmake is run and not on every build. So the dirty flag is not set if you build from a changed source but without rerunning cmake.

What do you think?  Feel free to discuss and criticise as you see fit.

cecilios:
Thank you for this contribution! I've been checking your proposal and I really like it. A very nice solution that improves current state. Thank you.

I've been checking also how this would affect my workflow and, as you said, nothing changes. Do you see any problem with the following workflow?:

- merge on GitHub
- if feel to make a release (mandatory if MAJOR or MINOR changes):
	- Decide the change in the MAJOR.MINOR version
	- add tag: always use the full version (0.17.0). The tag will create a new release when pushed.
	- update lomse_version.h. Use script, so that other files (i.e. debian/changelog) get updated.
	- commit and push to master -> this adds the tag and creates the release
	- go to the release and add some documentation describing the new release, add files, etc.
- else
	- Update lomse_version.h (use script, but only PATCH will be updated, as no tag changes)
	- commit and push to master

I'm new to GitHub and had still to learn how to use the Releases. But thanks to your comments I'm starting to have more clear ideas. Thanks again!

kolewu:
As for a tag, my suggestion is, to tag _some_ patch versions without any further ado. So I would think that it may be a good move to tag the current state as just 0.17.15 because it is now possible to build from a release package.  You don't need to do anything special, just tag and github will provide a download of the tagged version without the repository information that would be added when you checkout a tag with git.

Only do the whole story with release documentation when you increment the minor version.

So, yeah, you have described the workflow in a very concise way and I would only add to the _else_-part:
* maybe create a tag for the patch-version if you see fit, just to make it possible from time to time to get updated release packages without using git

cecilios:
Thank you! Got the idea. I will start soon using the Releases facility at GitHub.

Issue #35, opened 2016-03-11
move version information to cmake file


kolewu:
I have thought about the open point from #33: commit hash and dirty state is not updated without explicit run of cmake.

Do you see anything that speaks against removing the `include/lomse_version.h` and having the versions only in a cmake file?

The version information can be created like all the other cmake variables at cmake runtime. If you like the idea, I will prepare the necessary code changes. If you don't like it, I will try to change the cmake file to rebuild version strings with every change that affects them (like making the code dirty with respect to the repository).

cecilios:
As I see this:
- version information should be extracted from git
- But when building from a source package or if git is not installed in the machine, some static version information should be available and must be used. Currently, this is the role of lomse_version.h. 
- The library should have methods to provide:
    - the version info extracted from git or blank/empty string if not available at build time
    - the version info from the static variables

All these objectives are currently achieved by having the static information in lomse_version.h. But I don't see any problems for moving these static values to CMake. But in any case, what is the gain? Why do you think it is important to remove this info from a header file?

I don't understand your second point: "change the cmake file to rebuild version strings with every change that affects them (like making the code dirty with respect to the repository)". Do you mean, writing an script to be run after every commit so that it updates values in CMake file? Current script (scripts/update-version.sh) could be used for it. By the way I have to modify it because it was designed for using git tags in the form "major.minor" or "major.minor.0" but not current tag "0.17.20". With current script next version could be for instance "0.17.2"

kolewu:
I second your analysis with a small objection to the first: If the version is extracted from git **and** it's given in some file, there may be a difference that should at least be warned upon. If it's only given by git, it's not available in a source package without a special release process for building the source package. And a special release process makes it impossible to just use tags as release packages :disappointed: 

It's not important to remove the version information from a header file but if it's included in a cmake file it's not necessary to parse a header file to get it. So it's in the book of removing unnecessary code to reduce possible errors.

And to the second question: Imagine this workflow

- checkout the code from the repo
- run cmake
- run make

all versions are what we need and expect

but then

- make some changes in the code
- run make

the version still is the same as the one from the fresh checkout. But we expect a dirty in the version string. This is only added if cmake is run after changing anything. The same if you change the version file. And at least this version changed can be easily  managed by having the version information in the cmake files, because every change in a cmake file will trigger a cmake run by only running the normal build process. I'm not sure about how to achieve the same for changes in the git information, but will try to find out.

[...]

PR #36, opened 2016-03-18
Better versionnumbers

kolewu:
Here is my proposal for #35 .

With this change the version information is derived from the `build-version.cmake` file and -- if available -- repository information (commit hash and dirty flag). It only affects the dependent files when something changes, so no unnecessary builds, even if the version number in `build-version.cmake` is changed.

The second change parses the version number from `git describe` including the patch version and updates the appropriate files.


cecilios:
Great! Thank you. 
A question: why to create helper.sh instead of putting its code inside update-version.sh? Do you prefer to have functions apart just in case they can be useful in other scripts?

kolewu:
I put them into an extra file to make them testable. With the extra file you can do this very easily:

    for d in "0.17.20-0-g0d8f9a3" "0.17.20-2-g25fd5cc" "0.17" "error"
    do
        parseDescription "$d"
        echo
    done

It should be a real test, but to make sure the implementation works the same as the one before, it's sufficient.

cecilios:
Thank you. Merged.

cecilios:
Closes #35

Issue #68, opened 2017-09-11
MusicXML importer: assume some default values for clef


cecilios:
Some MusicXML files (see issue #66) do not specify the clef. These files are redenred without problems in other applications by making assumptions and taking defaults that, to my knowledge, are not in the MusicXML standard. 

Therefore, although Lomse is behaving correctly, the support for importing MusicXML could be improved by also adding defaults and heuristics for dealing with files with incomplete clef information.

PR #78, opened 2017-09-25
Fix issue #68: Assume some default value for clef in MusicXML importer

cecilios:
* Added class AutoClef for enclosing the algorithm for issue #68.
    * Added a few of helper methods in StaffObjsCursor.
    * Added unit tests for AutoClef class.


cecilios:
Lomse API documentation has been updated to include options for MusicXML importer and how to use them. See https://lenmus.github.io/lomse/classMusicXmlOptions.html

PR #126, opened 2018-04-09
Fixed compilation errors on Windows

hugbug:
This fixes compilation errors on Windows. I'm using Visual Studio 2015 but the changes are not specific to that compiler version.

## 1. Name conflicts
The following error occurs for many compilation units:
```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\thr/xtimec.h(24): error C2872: 'xtime': ambiguous symbol (compiling source file src\graphic_model\lomse_caret.cpp)
```
It happens due to simultaneous `using boost` and `using std`.
The fix was to not `using boost`. That was easy because the boost symbols are used with full qualifiers throughout the lomse code, except one place where I added boost names prefixes.

## 2. minizip/ioapi.c
Compilation fails at:
```C
void fill_fopen_filefunc (pzlib_filefunc_def)
  zlib_filefunc_def* pzlib_filefunc_def;
{
...
```
with
```
>packages\minizip\ioapi.c(223): error C2065: 'pzlib_filefunc_def': undeclared identifier
1>packages\minizip\ioapi.c(223): error C2182: 'fill_fopen_filefunc': illegal use of type 'void'
1>packages\minizip\ioapi.c(223): error C2365: 'fill_fopen_filefunc': redefinition; previous definition was 'function'
```
The error can be fixed by changing compiler options for individual file `ioapi.c` and setting option "Compile as" to "Compile as C Code" (instead of default "Compile as C++ Code". It wasn't obvious to find that option. A better more universal fix is to change that code line to be C++ compatible:
```C++
void fill_fopen_filefunc (zlib_filefunc_def* pzlib_filefunc_def)
{
```

That's it. Thanks for consdering this PR.

### DIFF
--- a/packages/minizip/ioapi.c
+++ b/packages/minizip/ioapi.c
@@ -220,8 +220,10 @@ static int ZCALLBACK ferror_file_func (voidpf opaque, voidpf stream)
     return ret;
 }
 
+// Changed function declaration for better compatibility with C++ compilers:
+//void fill_fopen_filefunc (pzlib_filefunc_def)
+//    zlib_filefunc_def* pzlib_filefunc_def;
 void fill_fopen_filefunc (zlib_filefunc_def* pzlib_filefunc_def)
-
 {
     pzlib_filefunc_def->zopen_file = fopen_file_func;
     pzlib_filefunc_def->zread_file = fread_file_func;
--- a/packages/minizip/ioapi.c
+++ b/packages/minizip/ioapi.c
@@ -220,8 +220,8 @@ static int ZCALLBACK ferror_file_func (voidpf opaque, voidpf stream)
     return ret;
 }
 
-void fill_fopen_filefunc (pzlib_filefunc_def)
-  zlib_filefunc_def* pzlib_filefunc_def;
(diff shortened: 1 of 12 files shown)


packages/minizip/ioapi.c
```diff
@@ -220,8 +220,8 @@ static int ZCALLBACK ferror_file_func (voidpf opaque, voidpf stream)
     return ret;
 }
 
-void fill_fopen_filefunc (pzlib_filefunc_def)
-  zlib_filefunc_def* pzlib_filefunc_def;
```
cecilios: This file is third party code, so we should not change it but request the change from minizip maintainers. Nevertheless, this going to take too much time, and I'm not sure if this is still maintained.. So I would suggest to comment out the removed line and include a comment for future reference, so that when comparing a future version of ioapi.c (probably never) with current one we will spot the change.

Please review and let me know what you think

cecilios: All the other changes looks good to me. Thank you!

hugbug:
Added a code comment as you requested.

cecilios:
Thank you!

Issue #97, opened 2017-11-03
issue about example_1_win32.cpp


jellyr:
m_rbuf_window.attach(m_bitmap.buf(),
		m_bitmap.width(),
		m_bitmap.height(),
		-m_bitmap.stride()
	);

why the third argument is -m_bitmap.stride().  

when i change to  m_bitmap.stride() , it work, but the bmp is wrong , why ?

cecilios:
Sorry, but I don't know why stride is passed as negative value in Win32 examples.

I do not use Windows. But as it is a popular choice I tried to prepare some examples for Windows. As rendering is based on agg code I just copied the Windows code from agg samples. I tested all this time ago, with Windows XP, and worked fine.

I noticed that for other platforms, agg code worked directly with native platform images. But for Windows the approach followed in agg samples was different: it defined a Bitmap class for interfacing the real Windows bitmap. But I didn't do test for using directly the Windows bitmaps, in line as it is done for the other platforms. 

If the example does not work with Windows 10, probably the Bitmap class defined in the example has to be reviewed due to some changes between WinXP and Win10.

I would suggest to study the examples for Qt and wxWidgets and try to do it in a similar way for Windows.

If you find a solution, I would appreciate if you could post it here, for the benefit of other users and for fixing the examples. Thank you.

jellyr:
finally, i get bmp work out on windows 10.
change 

```
m_rbuf_window.attach(m_bitmap.buf(),
	m_bitmap.width(),
	m_bitmap.height(),
	-m_bitmap.stride()
```
);
to
```
	m_rbuf_window.attach(m_bitmap.buf(),
		m_bitmap.width(),
		m_bitmap.height(),
		m_bitmap.stride()
	);
```


And add 
	m_bmp->bmiHeader.biHeight =  -abs(m_bmp->bmiHeader.biHeight);
before;
	if (dvc_width != bmp_width || dvc_height != bmp_height)
	{
		::SetStretchBltMode(h_dc, COLORONCOLOR);

because msdn says 
StretchDIBits creates a top-down image if the sign of the biHeight member of the BITMAPINFOHEADER structure for the DIB is negative. For a code example, see Sizing a JPEG or PNG Image.
https://msdn.microsoft.com/en-us/library/windows/desktop/dd145121(v=vs.85).aspx

cecilios:
Thanks for reporting a solution.
I will keep this issue open until I have time (and a Windows machine) for fixing the example and testing it.

Thanks!!

[...]

PR #127, opened 2018-04-09
Allow negative stride and fix Windows example

hugbug:
This is a follow up to #97. The PR consists of two commits:
1. one to allow negative stride;
2. a fix by @jellyr to windows example to use positive stride.

Any one of these commits fixes the issue with Windows example. Applied together they work too as well.

Tested on Windows 10 and XP.


cecilios:
Thank you!

Issue #125, opened 2018-04-06
std::vector: reserve() vs resize()


hugbug:
There are many places in the code where the size of vector is supposedly set before writing into the elements in a loop. The code uses `reserve()` for that. However reserve() changes capacity and not the size. As a result further attempts to access vector elements fail when compiling in debug mode (assertion failures).

I can imagine that when compiling in release mode (no bounds checks) and not using `size()` for vectors the code may accidentally work as needed.

Here are few examples:
- [StaffObjsCursor::initialize_clefs_keys_times](https://github.com/lenmus/lomse/blob/fd283a4a83e7aed3ab1b773cd608356f3fcf4931/src/graphic_model/layouters/lomse_staffobjs_cursor.cpp#L60);
- [PartsEngraver::create_group_engravers](https://github.com/lenmus/lomse/blob/40cfeaeaa00429eb6217943d0e55687036e047bf/src/graphic_model/engravers/lomse_instrument_engraver.cpp#L90);
- [ColumnsBuilder::determine_staves_vertical_position](https://github.com/lenmus/lomse/blob/fd283a4a83e7aed3ab1b773cd608356f3fcf4931/src/graphic_model/layouters/lomse_spacing_algorithm.cpp#L482).
- etc.

Just search the code for "reserve". At most places where "reserve" isn't followed by "assign" the function "resize" must be used instead (alternatively `reserve` + `push_back` instead of `operator[]`).

Furthermore at places where "assign" is called after "reserve" the "reserve" is superfluous since "assign" knows the size and can reserve capacity itself.

cecilios:
Many thanks for spotting this.

I used resize() to avoid reallocations as elements are inserted, followed by assign() to initialize the vector when filling the vector will not be done by push_back(). Unfortunately, as you spotted, in some cases I forgot the assign() part! And I didn't realize that reserve() is superfluous when followed by assign(), thanks.

I will review the code and try to fix this. In any case, for this or other issues, PRs are always welcome! ;-)

PR #128, opened 2018-04-09
#125: Fixed vector resizing

hugbug:
This is a fix to #125:
1) Using resize() instead of reserve() when need to access elements afterwards.
2) Removed unnecessary resize() before assign().


cecilios:
Thank you!

Issue #131, opened 2018-04-10
Trying to reduce or eliminate dependencies from other libraries


hugbug:
I'm working on a cross platform (free and open source) music application. Currently I compile and test it on Windows, macOS and iOS. Linux and Android will be supported later.

As there is no uniform build system for these platforms adding dependent libraries is a pain. The easiest would be to just add a new library into the project source code tree. I would do that with Lomse. I obviously can't do that for boost which is extremely heavyweight. Libpng, zlib and freetype - not sure about them either.

As an experiment I've tried to remove dependencies from other libraries:
- replaced **boost** with alternatives from standard library. More on this below.
- **zlib** - I only need to display scores, I don't need Lomse to save or load compressed files. I disabled zlib related parts by commenting them out. That was easy. A proper implementation would be to introduce a preprocessor define such as `LOMSE_DO_NOT_USE_ZLIB` and disabe related code parts with `ifdef`s. This define will be disabled by default (to enable all features) but can be activated when building Lomse (or when integrating it into source code tree of other project).
- **libpng** - I also don't need Lomse to read or save png-images. I've used the similar approach as with zlib and that was easy again.
- **freetype** - I didn't have much thoughts about this yet. Probably not avoidable but who knows. Maybe Lomse could have an abstract font drawer class and the user code could provide an implementation using user code classes (for example JUCE classes in my case).

The rest of this post is about boost.

## Boost
- `boost::thread`, `boost::mutex`, etc. replaced with their analogues from `std`:
   - `ScorePlayer` uses `timed_join` and `interrupt` which doesn't exist in standard library. I do not consider this to be a big issue because:
        * ScorePlayer already implements a proper signaling for interruption;
        * external thread termination is dangerous anyway and should be avoided;
   - `std::thread` requires **C++11**.
- `boost::format` replaced with `std::stringstream`:
   - `std::stringstream` isn't that comfortable to use but it works;
   - alternatively we can write an own printf-like function using `std::vsnprintf` (requires **C++11**).
- `boost::to_upper_copy` is used once. Replaced it with own implementation using `std::toupper`.
- `boost::variant` is used once in `AttribValue`. Using `union` instead.
- `date_time` module is used for two purposes:
  - obtaining current datetime and converting it into string, mostly for logging but also as time stamp in saved files. We can use `std::time` and `std::strftime` for that;
[...]

PR #132, opened 2018-04-11
Always use std::shared_ptr

hugbug:
This is a follow-up to #131.

`boost::shared_ptr` has been replaced with `std::shared_ptr`. For most systems this is rather a cosmetic change because `std::shared_ptr` was already used via `ifdef` in `lomse_basic.h`:

```C++
//if it is a C++11 compliant compiler use std shared_ptr, else use boost shared_ptr
#if __cplusplus > 199711L
    #include <memory>
    #define SharedPtr std::shared_ptr
#else
    #include <boost/shared_ptr.hpp>
    #define SharedPtr boost::shared_ptr
#endif
```

Visual C++ however was still using boost because Visual C++ defines `__cplusplus ` to `199711L` (although it supports most from C++11, C++14 and C++17).

With this PR Visual C++ starts using `std::shared_ptr` too, like other compilers.

### DIFF
--- a/include/lomse_basic.h
+++ b/include/lomse_basic.h
@@ -34,6 +34,7 @@
 
 #include <string>
 #include <vector>
+#include <memory>
 using namespace std;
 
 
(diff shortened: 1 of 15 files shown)


cecilios:
Thank you!

PR #133, opened 2018-04-12
Allow clang and other compilers

hugbug:
Currently compilation fails on any compiler other than GCC and MSVC. The cmake config explicitly checks for GCC and MSVC and defines symbols accordingly.

Then `lomse_logger.h` defines macros for logging based on compiler detection made by make. This macros rely on compiler specific  features and exist in two versions: for MSVC and for GCC. For other compilers the log macros are not defined at all. This cause compilation errors on other compilers, in particular **clang**.

I propose to remove explicit check for GCC and retain only check for MSVC. Let the code which was previously active only for GCC to be active on all compilers other that MSVC. This change is sufficient to compile on clang successfully. I wouldn't introduce an extra check for clang (it greatly mimics GCC).

Hopefully other compilers will work too. In any case these three compilers (GCC, MSVC, clang) are the most important nowadays and they all work after the fix.

### DIFF
--- a/build-options.cmake
+++ b/build-options.cmake
@@ -114,10 +114,8 @@ endif()
 # compiler
 if(MSVC)
     set( LOMSE_COMPILER_MSVC  "1")
-    set( LOMSE_COMPILER_GCC   "0")
-elseif(CMAKE_COMPILER_IS_GNUCC)
+else()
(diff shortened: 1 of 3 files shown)


cecilios:
Thanks for this PR

PR #134, opened 2018-04-12
Fixed crash in chord engraver layout_accidentals

hugbug:
During testing of example [musicxml-files](https://www.musicxml.com/music-in-musicxml/example-set/) encountered program crashes on most of the files.

The issue has been localized to an iterator incremented two times in a row without bounds check.

After the fix 17 from 18 files are read (and displayed) without crash. One file still causes a crash but I wasn't able to fix this. In a case you want to fix this yourself - this is file "ActorPreludeSample.musicxml", which is the biggest one (1 MB) in the set.

### DIFF
--- a/src/graphic_model/engravers/lomse_chord_engraver.cpp
+++ b/src/graphic_model/engravers/lomse_chord_engraver.cpp
@@ -342,8 +342,8 @@ void ChordEngraver::layout_accidentals()
             {
                 GmoShapeNotehead* pHead = (*itCur)->pNoteShape->get_notehead_shape();
                 shift_acc_if_confict_with_shape(pCurAcc, pHead);
+                ++itCur;
             }
-            ++itCur;
             if (itCur != m_notes.rend())
             {
                 GmoShapeNotehead* pHead = (*itCur)->pNoteShape->get_notehead_shape();


cecilios:
Interesting!

Before accepting a PR I always do a regression test in my machine and MusicXML samples are part of this, but didn't get any crash. I normally upload the results to the website. You can see the results after merging last PR at http://lenmus.org/en/lomse/regression#recordare

The file "ActorPreludeSample" always fails (but no crash) because it is an orchestral score that doesn't fit in default DIN A4 paper used by Lomse.

Probably this is due to a difference between Windows and Linux, and the bug was unnoticed in Linux. Anyway, I'm curious to see the results after applying your PR. Also I would like to check if there is any difference in current samples from the ones I downloaded time ago.

hugbug:
I tried these files in a test program based on example_1_win32.cpp. Maybe Lomse is configured there with other settings and that's why it processes the file differently (for example renders it regardless of canvas size)?

cecilios:
Findings:
- The crash in file "ActorPreludeSample" is a bug related to using single-system-view. This has been fixed. In my regression tests I use vertical-view. 
- I've updated the MusicXML samples I was using for tests. The ones I was using were older versions. Regression tests with the new versions do not crashes, either without your fix in this PR or with it. Therefore, the difference must be something related to std behaviour in the different platforms so that in Linux the bug was unnoticed.
- I have also added a fix so that the new extension 'musicxml' for MusicXML files is now accepted.

cecilios:
Thanks for this PR

PR #135, opened 2018-04-12
Fixed compilation error in minizip unit on macOS

hugbug:
Minizip code uses `ftello64`, `fseeko64`, `fopen64`. MacOS doesn't have them and compilation fails.

Clang produces warnings during compilation:
```
Implicit declaration of function 'fopen64' is invalid in C99
Implicit declaration of function 'ftello64' is invalid in C99
Implicit declaration of function 'fseeko64' is invalid in C99
```
and errors during linking:
```
Undefined symbols for architecture x86_64:
  "_fopen64", referenced from:
      _fopen64_file_func in ioapi.o
  "_fseeko64", referenced from:
      _fseek64_file_func in ioapi.o
  "_ftello64", referenced from:
      _ftell64_file_func in ioapi.o
```

As suggested on stackoverflow (also related to minizip) a solution is to define the symbols as:
```C++
#ifdef __APPLE__
    #define ftello64 ftello
    #define fseeko64 fseeko
    #define fopen64 fopen
#endif
```

That indeed helps. The question is only what is the best place for that code snippet? I've put it into minizip's file `ioapi.c` because this is the only unit needing them. If you think there is a better place for the snippet please let me know and I'll move it there (or you can do that yourself of course).

Thanks.

### DIFF
--- a/packages/minizip/ioapi.c
+++ b/packages/minizip/ioapi.c
@@ -16,6 +16,7 @@
 
 #include "ioapi.h"
 
+//Lomse. Fix compilation on macOS, where ftello64, fseeko64 and fopen64 are not defined.
 #ifdef __APPLE__
     #define ftello64 ftello
     #define fseeko64 fseeko
@@ -88,7 +89,7 @@ static int     ZCALLBACK ferror_file_func OF((voidpf opaque, voidpf stream));
 
 static voidpf ZCALLBACK fopen_file_func (voidpf opaque, const char* filename, int mode)
 {
-    (void)opaque;     //CSG. Fix for warning due to unused param
+    (void)opaque;     //Lomse. Fix for warning due to unused param
     FILE* file = NULL;
     const char* mode_fopen = NULL;
     if ((mode & ZLIB_FILEFUNC_MODE_READWRITEFILTER)==ZLIB_FILEFUNC_MODE_READ)
@@ -107,7 +108,7 @@ static voidpf ZCALLBACK fopen_file_func (voidpf opaque, const char* filename, in
 
 static voidpf ZCALLBACK fopen64_file_func (voidpf opaque, const void* filename, int mode)
 {
-    (void)opaque;     //CSG. Fix for warning due to unused param
+    (void)opaque;     //Lomse. Fix for warning due to unused param
 
     FILE* file = NULL;
     const char* mode_fopen = NULL;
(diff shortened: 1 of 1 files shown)


cecilios:
ioapi.c has already a lot of patches :-( so one more is not a problem, and for this case it its the best place. Please add, before your code, a comment line explaining the reason such as:
//Lomse. Fix compilation error in minizip unit on macOS

Also, add the prefix "Lomse. " to the comment in line 229 that you added the other day. And, if you don't mind, replace "CSG." (my name initials) by "Lomse." in several comments with changes I did time ago.

Although Git allows to find the changes I would prefer to have them easily identified in this file for future reference, just in case. Thanks in advance.

By the way, how do you build the lomse library? Do you use the provided CMakeLists.txt file for Windows? And for macOS?

hugbug:
Added comments as requested.

> By the way, how do you build the lomse library? Do you use the provided CMakeLists.txt file for Windows? And for macOS?

On Windows I had a great deal of troubles with CMake (never used it before). First it couldn't find compiler, then it couldn't find libraries etc. Eventually CMake has generated project file for Visual Studio but could not compile it on it's own. But at this point I had the project containing all needed files. I added required libraries and search paths directly in Visual Studio. I changed project type from library to executable and added the windows example into it. This way I can compile the example application and debug through Lomse files. I can't run unit tests on Windows yet.

On Mac the things work much better. CMake generates "Unix Makefile" without troubles and compiling works (after submitted fixes). The occurred errors was easy to fix. In another build directory I run CMake with target "Xcode" and it generated Xcode project, which I then opened in Xcode and could compile without issues (after applying fixes). In both "Unix Makefile" and "Xcode" targets I can compile library and testlib. In Xcode I can run and debug tests.

On Mac I have some test failures but that is worth a separate issue I guess.
```
FAILURE: 35 out of 1577 tests failed (86 failures).
Test time: 9.31 seconds.
```
Complete log is [here](https://pastebin.com/qY1XThk6).

cecilios:
Thanks for the changes and the info on building. I never tested the cmake file in Windows and, for sure, need fixes. 

Just in case, in relation to failures in the unit tests, I'm having since a month, a couple of random errors, always in two tests (see http://lenmus.org/en/lomse/unit-tests). They only happen randomly in my machine, never in Travis. In a quick look I didn't find the cause but had no time for investigating more.

I expect to merge this PR tomorrow morning.

cecilios:
All unit tests error messages such as "Line 112. Invalid ending number '1, 2'. <ending> ignored." are originated in `lomse_mxl_analyser.cpp`, line 2804, method  `bool mxl_is_valid_ending_number(const string& num)`. Looks like a problem with std::regex

cecilios:
Also all failures in tests such as MxlAnalyser_direction_words_04 are due to std::regex behaviour in mac

cecilios:
Also all problems related to volta_bracket seem to have its origin in method `bool mxl_is_valid_ending_number(const string& num)` in `lomse_mxl_analyser.cpp` and related to sdt::regex.

PR #137, opened 2018-04-13
Support newer versions of UnitTest++ (1.5+)

hugbug:
The test runner of **testlib** [contains](https://github.com/lenmus/lomse/blob/0.24.0/src/tests/lomse_the_test_runner.cpp#L89):
```C++
  Test* q = p;
  p = p->next;
  q->next = nullptr;
```

Here member `next` of class `UnitTest::Test` is accessed. This class member seems to be not meant for usage outside of UnitTest++ internal implementation. Otherwise it's unexplainable why the authors of UnitTest++ have renamed it to `m_nextTest` in version **1.5** of the library.

This change breaks compilation of **testlib** if UnitTest++ version 1.5 or newer is used.

Most Linux distribution have already updated to UnitTest++ 2.0 or 1.6 - search for "libunittest" and "unittest-cpp" on [pkgs.org](https://pkgs.org) (direct links to search results don't work unfortunately).

The easiest "fix" would be to simply change `p->next` to `p->m_nextTest` in the Lomse test runner. That would break library compatibility with older systems though.

UnitTest++ doesn't declare any version constants which we could use in `ifdef`.

## Proposed solution
In this PR the problem is solved by detecting UnitTest++ version during CMake configuration step. For that we create a tiny temporary program containing code which tries to access member `next` of class `Test`. We try to compile that program. If compilation succeeds we know that we have UnitTest++ v1.4 or older. If compilation fails we assume UnitTest++ is 1.5 or newer.

Depending on the result of compilation attempt we write option `LOMSE_UNITTEST_OLD` into `lomse_config.h` and use it later in the test runner:
```C++
#if (LOMSE_UNITTEST_OLD == 1)
    p = p->next;
    q->next = nullptr;
#else
    p = p->m_nextTest;
    q->m_nextTest = nullptr;
#endif
```

### Caveat regarding lomse_config.h
Config file `lomse_config.h` is created when executing `build-options.cmake`, where configuration options such as disabling/enabling of unit tests are initialised. Therefore we cannot test UnitTest++ library version before `build-options.cmake` is executed but afterwards the config file `lomse_config.h` is already written and symbols defined later do not go into it.

To solve this dilemma I've copied the command for creation of `lomse_config.h` (located at the end of `build-options.cmake`) to the end of `CMakeLists.txt`. The first call (from `build-options.cmake`) isn't necessary actually anymore but I kept it so that we can consider the second call (from `CMakeLists.txt`) to do update of the config and therefore legitimate.

### Results
Tested this on Mac (UnitTest++ 2.0), Windows (UnitTest++ 2.0) and on Linux (Travis with UnitTest++ 1.4).


CMakeLists.txt
```diff
@@ -170,6 +171,22 @@ if (LOMSE_BUILD_TESTS)
...
+            "int main() { UnitTest::Test* test; test->next = 0; return 0; }")
+        try_compile(LOMSE_UNITTEST_HAS_NEXT ${CMAKE_CURRENT_BINARY_DIR} "${CMAKE_CURRENT_BINARY_DIR}/unittest++-version-test.cpp"
+            CMAKE_FLAGS -DINCLUDE_DIRECTORIES=${UNITTEST++_INCLUDE_DIR})
+        file(REMOVE "${CMAKE_CURRENT_BINARY_DIR}/unittest++-version-test.cpp")
+        if(LOMSE_UNITTEST_HAS_NEXT)
+            message(STATUS "Detecting UnitTest++ version - done (<1.5)")
+            set(LOMSE_UNITTEST_OLD "1")
+        else()
+            message(STATUS "Detecting UnitTest++ version - done (>=1.5)")
+            set(LOMSE_UNITTEST_OLD "1")
```
cecilios: Shouldn't be  `set(LOMSE_UNITTEST_OLD "0")` ??

hugbug: Of course! Forgot to set it back to "0" after doing simulation tests in a final clean PR.

Sorry for that.

Submitted the fix.

cecilios:
Thanks. A clever solution!

PR #139, opened 2018-04-13
Fixed test failures for invalid font size

hugbug:
This PR fixes:

>All unit tests error messages such as "Line 0. Invalid size '10pt'. Replaced by '12'." are originated in lomse_ldp_analyser.cpp, line 361, method float get_font_size_value(). I do not understand why. Received parameter should be something as "10pt".

It attempts to read from `istream` a text such as "10pt" whereas the intention was to read the text with stripped suffix such as "10".

It seems libstdc++ (used on Linux) reads until the first non-digit character and succeed (and therefore the bug went unnoticed). With libc++ (clang) any non-digit character causes reading failure. From my understanding of description of `operator >>` the clang's behaviour is the correct one.


cecilios:
Thank you

PR #142, opened 2018-04-16
Replacing boost::format with std::stringstream

hugbug:
As part of eliminating dependency from `boost` we replace `boost::format` with an alternative solution which relies on standard library only.

I've made two variants of the replacement:
- the one which uses `std::stringstream`;
- and another one with an own `printf`-like function;

I'm posting both variants as separate PRs and let you choose which one you like more.

In this PR we use `std::stringstream`.

The replacement looks like this:
#### was:
```C++
    LOMSE_LOG_TRACE(Logger::k_layout, str(boost::format(
        "Trying to create the EngroutersCreator for Imo id %d %s")
        % pImo->get_id() % pImo->get_name() ));
```

#### became:
```C++
    stringstream ss;
    ss << "Trying to create the EngroutersCreator for Imo id " <<
        pImo->get_id() << " " << pImo->get_name();
    LOMSE_LOG_TRACE(Logger::k_layout, ss.str());
```

It's a little bit more verbose but manageable. However there is one issue. `LOMSE_LOG_DEBUG` and `LOMSE_LOG_TRACE` are macros which in debug builds expand to calls `logger.log_debug()` and `logger.log_trace()`. In release builds these macros expand to empty code.

Therefore in release mode:
- variant with `boost::format` produces no code.
- variant with `stringstream` instantiates a stringstream object and writes into it, the result isn't used. I doubt compilers can eliminate instantiation and writing into stream here.

To avoid this unnecessary runtime overhead in release builds we should write:
```C++
#if (LOMSE_ENABLE_DEBUG_LOGS == 1)
    stringstream ss;
    ss << "Trying to create the EngroutersCreator for Imo id " <<
        pImo->get_id() << " " << pImo->get_name();
    LOMSE_LOG_TRACE(Logger::k_layout, ss.str());
#endif
```

This is considerably more verbose than the original boost-version.

These additional `#if (LOMSE_ENABLE_DEBUG_LOGS == 1)` are not included in this PR but they probably should be added if you choose that alternative.

The second approach (own printf-like function, for which I'll post another PR) works similar to `boost::format` version and doesn't produce code in release builds.

### DIFF
--- a/include/lomse_logger.h
+++ b/include/lomse_logger.h
@@ -37,10 +37,8 @@
 #include <iomanip>
 #include <fstream>
 #include <string>
+#include <sstream>
 using namespace std;
-
-//other
-#include <boost/format.hpp>
 
 namespace lomse
 {
--- a/src/file_system/lomse_zip_stream.cpp
+++ b/src/file_system/lomse_zip_stream.cpp
@@ -35,6 +35,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <cstring>
+#include <algorithm>  //min()
 using namespace std;
 
 namespace lomse
--- a/src/graphic_model/engravers/lomse_engrouters.cpp
+++ b/src/graphic_model/engravers/lomse_engrouters.cpp
@@ -40,7 +40,6 @@
 #include "lomse_logger.h"
 
 //other
-#include <boost/format.hpp>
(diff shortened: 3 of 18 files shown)


cecilios:
Finally, implemented version in PR #144, as discussed in #143. Thank you.

PR #143, opened 2018-04-16
Replacing boost::format with own format function

hugbug:
This is an alternative solution for replacing `boost::format`.
Related issue: #142.

Our own format function:
```C++
    std::string format(const char* fmtstr, ...);
```

I've placed it into `lomse_logger.h/.cpp`. Is that a proper place?

Usage:

#### was:
```C++
    LOMSE_LOG_TRACE(Logger::k_layout, str(boost::format(
        "Trying to create the EngroutersCreator for Imo id %d %s")
        % pImo->get_id() % pImo->get_name() ));
```

#### became:
```C++
    LOMSE_LOG_TRACE(Logger::k_layout, format(
        "Trying to create the EngroutersCreator for Imo id %d %s",
        pImo->get_id(), pImo->get_name().c_str() ));
```

Function `printf` is traditionally considered unsafe because passing parameters of incorrect types (as expected in format string) can cause crashes. However GCC and clang check parameters of printf-family functions and print warnings like this:
```
src/graphic_model/engravers/lomse_engrouters.cpp:92:13: Format specifies type 'int' but the argument has type 'const char *'
       LOMSE_LOG_TRACE(Logger::k_layout, format(
            "Trying to create the EngroutersCreator for Imo id %d %s",
            "test error", pImo->get_name().c_str() ));
            ^~~~~~~~~~~~~~
```

Even better these compilers expand the checks to user defined functions by marking functions with a special attribute:
```C++
    std::string format(const char* fmtstr, ...) __attribute__((format(printf, 1, 2)));
```

All potential issues with formatting arguments are catched at compile time.

### DIFF
--- a/include/lomse_logger.h
+++ b/include/lomse_logger.h
@@ -39,9 +39,6 @@
 #include <string>
 using namespace std;
 
-//other
-#include <boost/format.hpp>
-
 namespace lomse
 {
 
@@ -71,6 +68,14 @@ extern ofstream dbgLogger;
         #define LOMSE_LOG_TRACE(area, msg)
     #endif
 
+#endif
+
+#ifdef __GNUC__
+// GCC and Clang support error detection of format arguments in compile time
+std::string format(const char* fmtstr, ...) __attribute__((format(printf, 1, 2)));
+#else
+// This compiler does not support error detection of format arguments in compile time
+std::string format(const char* fmtstr, ...);
 #endif
 
 
--- a/src/file_system/lomse_zip_stream.cpp
+++ b/src/file_system/lomse_zip_stream.cpp
@@ -35,6 +35,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <cstring>
+#include <algorithm> // min()
 using namespace std;
 
 namespace lomse
--- a/src/graphic_model/engravers/lomse_engrouters.cpp
+++ b/src/graphic_model/engravers/lomse_engrouters.cpp
@@ -40,7 +40,6 @@
 #include "lomse_logger.h"
 
 //other
-#include <boost/format.hpp>
 #include "utf8.h"
 
 namespace lomse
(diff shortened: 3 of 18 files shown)


cecilios:
After considering your two approaches on this issue, my current opinion is:

- I'm still inclined to use iostreams to format strings. It is the modern way of doing it and it is safe. It is more verbose but it is not clear if this is a drawback or and advantage (more legible code and easy to understand by young programmers not used to printf details).

- For `LOMSE_LOG_XXXX` macros it its preferable a solution more compact and that do not generate code in release builds. A solution using an own format function is better.

Therefore I would suggest the following approach:

- define additional logger macros accepting a format string and a list of arguments, instead of a message; Example:
```
#define LOMSE_LOG_ERROR(msg) logger.log_error(__FILE__,__LINE__,__FUNCTION__, msg);
#define LOMSE_LOG_ERROR(const char* fmtstr, ...) \
	logger.log_error(__FILE__,__LINE__,__FUNCTION__, const char* fmtstr, __VA_ARGS__);
```

- move your new format funtion to the logger class, as a private member method

- use stringstream for all other uses in lomse code.

What is your opinion on all this? (your two proposed solutions and this third one). Please, do not code the third solution until we agree on the best approach!

hugbug:
Thanks for taking your time for all the PRs. I appreciate it.

I'm not sure about iostreams to be considered modern. That invention of C++ is like 30 years old but wasn't adopted by newer languages: Java, Python, Ruby, Swift, Kotlin. I believe none of theme have a similar concept but all of them have format capabilities (either as functions or even as built-in language features). Therefore I think for new programmers a format function is easier to comprehend than iostreams.

Anyway the most places where formatting is necessary in Lomse are related to logging. I'm finding the third alternative very good and would be happy to implement it. It has an advantage of keeping printf-like function an **internal detail**.

There is a [C++ standard proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0645r0.html) for format feature. That proposal is based on existing library [fmt](https://github.com/fmtlib/fmt) (over 3600 stars on GitHub). The implementation consists of two files (format.h and format.cc) which needs to be added to the project. Would you possibly like that alternative better than others? I haven't tried it myself yet though.

I'll be happy with any approach you choose, as long as it helps to loose dependency from boost and doesn't add other dependencies (adding two external files to the project doesn't count).

cecilios:
Thanks for your comments.

Then I would vote for the third approach. Using the fmt library, even in source code, will add a dependency from it, as we will be forced to follow that project to be informed about changes and bugs. As you said, lomse uses format mainly for logging and this will be solved by an specific solution. Therefore, no need for more complexity for all other cases; use iostreams.

Please, if you encounter any difficulty in coding this third solution, let me know. Do not spend too much time on this as there are alternatives.

cecilios:
Finally, merged version implemented in #144. Thank you.

PR #146, opened 2018-04-18
Replaced boost::variant with union class

hugbug:
In this PR `boost::variant` is replaced with **union class**.

Along with plain types (int, float) the union needs to hold objects too (string, Color). This requires using of **placement new** and **explicit destructor calls** as explained in [Union declaration](http://en.cppreference.com/w/cpp/language/union) (cppreference.com).


cecilios:
Thank you!

PR #152, opened 2018-04-20
Fixed tests "Dynamic_GeneratesRequest" failing in MSVC

hugbug:
In the following example:
```C++
TEST_FIXTURE(LdpAnalyserTestFixture, Dynamic_GeneratesRequest)
{
    void* p1 = this;
    void* p2 = static_cast<LdpAnalyserTestFixture*>(this);
```
In **clang** and **GCC** `p1 == p2` but not in **MSVC**, at least not for Fixture classes.

Therefore storing `this` in `set_request_callback` for the class inherited from `LdpAnalyserTestFixture*` and then type casting it to `LdpAnalyserTestFixture*` in `wrapper_lomse_request` causes access into wrong memory:
```
static void wrapper_lomse_request(void* pThis, Request* pRequest)
{
    static_cast<LdpAnalyserTestFixture*>(pThis)->on_lomse_request(pRequest);
```

This PR fixes the issue by passing to `set_request_callback` the object of the type needed in that callback (type-casted).

### DIFF
--- a/src/tests/lomse_test_ldp_analyser.cpp
+++ b/src/tests/lomse_test_ldp_analyser.cpp
@@ -10062,7 +10062,7 @@ SUITE(LdpAnalyserTest)
     TEST_FIXTURE(LdpAnalyserTestFixture, Dynamic_GeneratesRequest)
     {
         LomseDoorway* pDoorway = m_libraryScope.platform_interface();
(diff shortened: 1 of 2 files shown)


cecilios:
Thank you. Curious bug!

PR #159, opened 2018-04-26
Make dependencies from zlib and pnglib optional

hugbug:
As previously discussed in #131 this PR implements the possibility to build Lomse without explicit dependency from zlib and pnglib.

**Note:** Freetype may still require zlib but that's regulated outside of Lomse build process. In particular on Windows official builds of Freetype don't require separate zlib (Freetype.dll probably has zlib statically integrated).

By default zlib and pnglib are used in Lomse like before and all features are enabled. To build without pnglib and/or zlib there are two CMake options (both `ON` by default):
 - LOMSE_ENABLE_COMPRESSION;
 - LOMSE_ENABLE_PNG.

In the following example we run CMake providing these two options:
```shell
cmake -G "Unix Makefiles" -DLOMSE_ENABLE_PNG=OFF -DLOMSE_ENABLE_COMPRESSION=OFF  ../lomse
```

**Note:** Pnglib can be deactivated separately but deactivating of zlib also requires deactivating of pnglib because pnglib depends on zlib. Therefore `LOMSE_ENABLE_COMPRESSION=OFF` also requires `LOMSE_ENABLE_PNG=OFF`.

Status of the new options is printed during running CMake similar to other options:
```
-- Compatibility for LDP v1.5 = ON
-- Enable compressed formats = OFF
-- Enable png format = OFF
```

`lomse_config.h` contains two new preprocessor defines:
```C++
// Enable compressed formats (requires zlib)
#define LOMSE_ENABLE_COMPRESSION    @LOMSE_ENABLE_COMPRESSION@

// Enable png format (requires pnglib and zlib)
#define LOMSE_ENABLE_PNG    @LOMSE_ENABLE_PNG@
```

Code parts depending on zlib or pnglib are guarded using these defines, for example:
```C++
#if (LOMSE_ENABLE_PNG == 1)
// PngImageDecoder: knows how to read a PNG image
class PngImageDecoder : public ImageDecoder
... 
};
#endif // LOMSE_ENABLE_PNG
```

### Testing
Tested on Mac (Clang) with default options and with deactivated libraries. All compiles successfully, all tests pass.

My tests on Linux were limited to Travis build with default options (zlib and pnglib are both enabled).

Also tested on Windows, library compiles successfully. As mentioned in other PR testlib application still fails to link due to some misconfiguration. That's a separate issue not related to current PR. I don't know how to fix this yet, that needs a separate investigation.

### Extras
I've found that library config file `liblomse.pc.cmake` was still referencing boost libraries. I've fixed this within this PR but in a separate commit. I thought it wasn't worth a separate PR.

---
This PR has option **Allow edits from maintainers** activated. Please feel free to commit any necessary modifications or tell me if you want me to change something.

Thanks for very quick reviewing and merging of my PRs!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,25 @@ include( ${LOMSE_ROOT_DIR}/build-options.cmake )
 #
 ###############################################################################
 
+#dependencies for building and for pkg-config file (intitially empty but will be enhanced)
+set(LOMSE_REQUIRES "")
+set(LOMSE_DEPENDENCIES "")
+
+# include directories to be installed
+set( INCLUDES_LOMSE  ${LOMSE_ROOT_DIR}/include )
+set( INCLUDES_AGG  ${LOMSE_ROOT_DIR}/src/agg/include )
+set( INCLUDES_UTFCPP  ${LOMSE_ROOT_DIR}/packages/utfcpp )
+set( INCLUDES_MINIZIP  ${LOMSE_ROOT_DIR}/packages/minizip )
+set( INCLUDES_PUGIXML  ${LOMSE_ROOT_DIR}/packages )
+include_directories(
+    ${INCLUDES_LOMSE}
+    ${INCLUDES_AGG}
+    ${LOMSE_ROOT_DIR}/src/agg/font_freetype
+    ${INCLUDES_UTFCPP}
+    ${INCLUDES_PUGIXML}
+)
+
+
 # Check for UnitTest++. Required for unit test
 if (LOMSE_BUILD_TESTS)
     find_package(UnitTest++)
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -49,38 +49,75 @@
...
+# LOMSE_ENABLE_COMPRESSION   (Default value: ON)
+#	Build with support for compressed files. This requires zlib library.
+#	By setting this to OFF Lomse will not be able to deal with compressed files,
+#	such as MusicXML files in compressed format, or with png images embedded in
+#	the documents. But Lomse will not fail in those cases, just will log error
+#	messages. For instance:
+#		cmake -G "Unix Makefiles" -DLOMSE_ENABLE_COMPRESSION=OFF ....
+#
+#	**Important**: deactivating compression will also deactivate support for
+#	PNG images because libpng depends on zlib.
```
hugbug: The description differs from current implementation: trying to disable compression without also disabling png has no effect. If png is active then compression is automatically activated.

If you want it the way you describe it that can be easily achieved but requires small modifications to CMake configuration.

cecilios: Thanks. I will fix the description.

I've tested the PR and it works as it is now. But I'm considering to remove the crash (throw) when the library has been built without zlib support and it is requested to open a compressed file. I think it is better not to throw an exception but to log an error message and fail opening the file (this will report a 'not supported format' error, and will not cause a crash in the user application).

hugbug: I saw exceptions in other places and thought it's the way to report errors. It's certainly not good to throw exception in one case but report errors via empty objects or using error codes in others. Consistency of behaviour is important.

Please make it the way you like as you of course know design choices of your project better.

cecilios: I started with this time ago. At start, I always threw exceptions. But after starting to use the library in my application I realize that when feasible and convenient it is better to log errors and do not cause crashes. There are still many points in which the library throw exceptions, but I didn't have time for reconsidering them and do not cause problems in applications. Most of them are at places in which there must have been a very bad failure in logic or program design.

cecilios:
Thank you!

PR #161, opened 2018-04-29
Fix issues detected with GCC 7

cecilios:
Most of the issues were just cosmetic changes. But a couple of them were real bugs (using operator ~ instead of !).

### DIFF
--- a/include/lomse_tree.h
+++ b/include/lomse_tree.h
@@ -99,7 +99,9 @@ public:
             depth_first_iterator& operator ++()
             {
                 if (m_currentNode == nullptr)
+                {
                     return *this;   //we are at end and will remain there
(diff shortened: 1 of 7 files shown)


hugbug:
When GCC reports misleading indentation after if statements that's because the source files contain mixed indentation with tabs and spaces. By default GCC assumes tab is equal to 8 spaces. However in Lomse sources tab is equal to 4 spaces.

When GCC reports such warning you should check if if-statement and the next line are both indented with the same indentation character. You don't have to add curly braces.

Another possibility is to use special GCC option `-ftabstop=4`.

You may also want to check warnings from clang. There are some interesting reports about overridden virtual functions with changed parameter list.

PR #170, opened 2018-06-14
ScorePlayer thread start synchronisation

hugbug:
`ScorePlayer` has member `m_pThread` pointing to playback thread. It is initialised in `ScorePlayer::play_segment`:
```C++
m_pThread = LOMSE_NEW SoundThread(&ScorePlayer::thread_main, this,
```

and then used in `ScorePlayer::do_play`:
```C++
if (!m_pMidi || !m_pThread)
{
      LOMSE_LOG_DEBUG(Logger::k_score_player, "<< Enter. No Midi or no thread. << Exit");
      return;
}
```

Depending on how thread scheduler on OS works the new thread may become active before the first thread continues execution. In this case `m_pThread` may be not yet initialised when it is checked in `do_play`, causing do_play to immediately terminate.

This happens pretty often on Windows causing ScorePlayerTest to fail. In theory this may happen on other OSes too but I didn't see this test failed on Linux or macOS yet.

The proposed fix uses mutex to ensure the main function of the new thread is executed after `m_pThread` is initialised.

### DIFF
--- a/include/lomse_score_player.h
+++ b/include/lomse_score_player.h
@@ -86,6 +86,7 @@ class ScorePlayer
 protected:
     LibraryScope&       m_libScope;
     SoundThread*        m_pThread;      //execution thread
+    std::mutex          m_startMutex;   //mutex so synchronize thread start
     MidiServerBase*     m_pMidi;        //MIDI server to receive MIDI events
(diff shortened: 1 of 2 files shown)


cecilios:
Hi, thanks for this. It looks like a reasonable change.

PR #185, opened 2018-07-25
Changes for rendering measure numbers

cecilios:
Measure numbers are now rendered. Options for controlling measures numbering have been added. MusicXML importer has been modified for importing measure numbering options.

In addition, some changes related to this:
- Define a new barline type for an alternative rendering for double repetition barlines. It uses two thick barlines instead of thin barlines (E.Gould, p.234).
- A couple of fixes in MusicXML importer, to eliminate some error messages and to fix a bug in measure numbers import.
- Style "Measure numbers" is now automatically added to the scores. By changing its settings the appearance of measure numbers can be customized.
- Defined a new barline type: alternative rendering for double repetition, using two thick barlines instead of thin barlines (E.Gould, p.234).


hugbug:
Thank you for this very welcome feature.
Works excellent for me!

sensibol:
Thank you.

PR #188, opened 2018-09-06
MusicXML: import metronome directions

cecilios:
This PR is, mainly, for supporting MusicXML metronome marks.  It improves the MusicXML support for importing the most common metronome construct used in most MusicXML files. But it does not add support for the alternative construct, used only for more complex metronome specifications. To support it I need to find sample MusicXML files using it.

This PR also includes a few additional changes:

- A long time ago, when metronome marks were supported in LDP, they were modelled by _ImoMetronomeMark_ objects, derived from _ImoStaffObj_. This decision was not the best one and several facts point to the need to reconsider it and to model _ImoMetronomeMark_ as an _ImoAuxObj_ attached to an _ImoDirection_. This PR does it.

- _ImoDirection_ was introduced time ago when importing things from MusicXML, but now there was a conflict with _ImoSpacer_ as, somehow, both do the same function. Therefore, _ImoSpacer_ has been removed and replaced by _ImoDirection_.

In addition, a few minor unrelated changes are included in this PR, such as removing compiler warnings due to logger class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ##### COMPATIBLE CHANGES
 
+- MusicXML importer now accepts metronome directions.
 - The algorithm for auto-scroll has been improved.
 - During playback, when the time signature changes, metronome clicks
   interval is now adjusted for maintaining notes duration equivalence.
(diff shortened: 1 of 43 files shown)


sensibol:
Thank you.

PR #211, opened 2019-09-10
Add methods to create/remove fragment marks on the score

cecilios:


cecilios:
PR #211 is an initial implementation. Finally, the class name is `FragmentMark`.

For now, the implementation is reduced to the necessary minimum and current implementation only deals with line marks. The idea is that you can start using it and detecting uncovered needs, and, meanwhile, I will continue developing the drawing of the other mark types (brackets, etc.).
 
 The creation methods are reduced to those providing timepos. I don't like the idea of offering additional alternatives for positioning, such as measure and measure,beat. Timepos is the basic and all the others requires transforming the arguments into a timepos. So for now, I would prefer this simple API. Using the other is just doing:

```C++
// note/rest position for measure, beat
TimeUnits timepos = ScoreAlgorithms::get_time_pos(measure,beat);
pInteractor->add_fragment_mark_at_note_rest(timepos);

// barline for measure iMeasure
TimeUnits timepos = ScoreAlgorithms::get_time_pos(iMeasure+1, 0);
pInteractor->add_fragment_mark_at_barline(timepos);

```

 
*Re-positioning the marker*

Unless really needed, I will not implement methods for re-positioning the marker. For different reasons, its is simpler to delete and re-create the marker at the new desired position than to implemet methods for moving the mark. In practice the difference is minimal:


```C++
// if repositioning is allowed
mark->move_to_note_rest(TimeUnits timepos);

// in current situation
pInteractor->remove_mark(mark);
mark = pInteractor->add_fragment_mark_at_note_rest(timepos);

```


*Using logical units or Tenths*

Current implementation uses logical units (cents of one millimeter) for specifying line thickness and extra height. But I am going to change this to use tenths. For now, I have preferred to commit the current code, so that you can start using and testing it. But please note that this will be changed.


Now I will update the documentation and then I will continue developing FragmentMark.

Issue #224, opened 2020-04-14
In some scores right barline is not justified


cecilios:
For instance:
![224-justification](https://user-images.githubusercontent.com/5238679/79211571-b5bb8f80-7e46-11ea-9c7f-90abb362f89e.png)

Justification problems seem to be related to lyrics. Compare test scores 50021 & 50021b.

There is one strange case without lyrics, test score 00110.

PR #225, opened 2020-04-14
Cumulative changes

### LINKED ISSUE #102: Select and use platform installed fonts
A solution for using platform installed fonts is required.

Currently, Lomse only uses the fonts distributed with liblomse package. This causes different problems:

- No proper text rendering when the score is using languages not supported in current fonts used by lomse (i.e. Japanese or Korean).

- In MusicXML files the font-name style attribute recommends fonts by name. Lomse should be able to select the most suitable font from installed fonts.

- Explicit font name information is not enough. In many cases there are different font files for each font name: i.e. one for normal font, another for bold and a third one for italic.

- As Lomse is platform independent, the solution must not force to add if preprocessor directives to select the platform and to add platform dependent code. But in any case, platform dependent code is needed. To solve this, some ideas:
    - Transfer the responsibility to the user application, that will inform lomse about font paths.
    - As the code for previous solution will be common to all applications in a given platform, this code could be packaged in platform dependent lomse modules for dealing with fonts. I will code only the module for Linux. The modules for all other platforms will have to be coded by contributors.
    - Enable both solutions. When there is no lomse fonts module for a platform, the user could simply inform lomse about fonts paths.

### LINKED ISSUE #223: Crash with certain musicxml files
I have two musicxml-files which cause crashes in the Lomse code.

Both files were created with MuseScore via importing of MIDI-files. Therefore these score files are far from what a living soul would make ;)

In my somewhat older version of Lomse (last updated Sep. 2019) I was able to fix the issue with [this two line commit](https://github.com/hugbug/conpianist/commit/60def42b7dd48745ceff4cdbdbf6aee7d8ada25f).

Interesting, the current version of Lomse has other code places causing troubles with these two score files. The following commit contains other fixes, needed to load and display the score without crashes in current Lomse version - [commit with more fixes](https://github.com/hugbug/conpianist/commit/f600be44986c658fceee724d2c2b1f0ca128055b).

Most likely the changes I made are not a proper way to fix the underlying issues. Therefore I'm not sending a PR.

## Files causing troubles
[Murka.musicxml.zip](https://github.com/lenmus/lomse/files/4464651/Murka.musicxml.zip)
[Can You Feel The Love Tonight-017.musicxml.zip](https://github.com/lenmus/lomse/files/4464652/Can.You.Feel.The.Love.Tonight-017.musicxml.zip)

## More info
I can provide stack traces in a case you can't reproduce the crashes.

cecilios:
This PR:
- Fixes #223 (Crash with certain musicxml files).
- Fixes #102 (Select and use platform installed fonts). Now lomse is not totally platform independent. For Linux (and macOS), current approach is to use _fonconfig_ library to determine font file to use and do font substitutions. But this creates a dependency from _fontconfig_ library. I have not tested in macOS. For Windows, code has been written for finding fonts in Windows fonts folder and making some substitutions. For other platforms, the old solution (fonts must be provided by user application) is applied.
- Logs destination folder. Lomse logs were polluting the system. This has been fixed and now log files are created at user home folder.
- Slurs. Some initial code for improving slurs layout to avoid collisions with other notations is now in place. 
- Other: Fix bug in anacrusis time computation


cecilios:
Travis CI passed.
https://travis-ci.org/github/lenmus/lomse
Thay are not displyaed !!!  Why?

cecilios:
Hi, thanks for this. Your suggestion is a reasonable change. I'm currently working on issue #224 but hope to finish this in two or three days. I will look at this as soon as possible. Could you please open an issue? Just copy your post. Thank you.

The rendering order is: 1st paper page (blank sheet), music for the page, 2nd paper page, music for the page ..., and finally the visual effects (e.g. markers) for displayed pages. This explains the visual appearance of the screenshots.

My first impression is that it is not much work to trim markers to paper height, until a better solution (auto-scaling?) can be coded.

Issue #235, opened 2020-05-08
Compiling in Windows using MinGW


de5car7es:
Hi

Having hard time compiling using MinGW.
First. generating makefiles had problem with if(CMAKE_NO_STD_NAMESPACE), for some reason this check fails, although I do have std. Anyway I've commented this out and created the makefiles.

The problem now is that make fails on\libs\Lomse\include/lomse_events_dispatcher.h:49:14: error: 'thread' in namespace 'std' does not name a type  49 | typedef std::thread EventsThread;

and some other similar errors on mutax.

I have installed mingw32 pthreads but it doesnt work.

Were you able to successfully compile it using MinGW ?

cecilios:
No, sorry. I never compiled the code using cmake in Windows, neither using windows tools nor MinGW. I only build the library in Linux and cmake buils the makefile withou problems. But in windows I use lomse including its sources in my app. LenMus, therefore I never tried cmake there. Probably it needs some fixes to build in Windows. The cmake file for lenmus, at https://github.com/lenmus/lenmus/blob/master/CMakeLists.txt builds the app using lomse sources in Windows under the ubuntu subsystem.

cecilios:
### Lomse build instructions for Windows 10

PR #236 partially fixes cmake script for building in Windows, as I explain here.

To build in Windows, open a Command Window appropiate for x64 build: _"x64 Native Tools Command Prompt"_

Create build directory:
```
cd path\to\lomse
rmdir build-area /S /Q
mkdir build-area && cd build-area
```
    
Create makefile and build:

- For building the static library:

```
cmake -G "Ninja" ..
ninja
```

You will see many warnings, mainly complaining about usafe function, such as:
> 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS.    

There are also a couple of other warnings that I have to fix.

The unit tests runner program `testlib.exe` is built and is linked against zlib.lib and libpng.lib but when executed, the testlib.exe program complains of not finding `zlib.dll` and `libpng.dll` (!!!???). Copying these DLLs to build folder solves the problem and the tests run. There are four failures related to Chinese font.

To install the library and the header files it is required to have admin privileges. To do it, open as admin a _"x64 Native Tools Command Prompt"_. Then:

```
cd path\to\lomse\build-area
ninja install
```


- For building the shared library:

```
[...]

PR #239, opened 2020-05-20
Improve MSVC compiler detection

cecilios:
Should fix https://github.com/lenmus/lomse/issues/235#issuecomment-631240230 by not setting LOMSE_COMPILER_MSVC to true. Not tested in Windows10 neither directly nor with MinGW.


cecilios:
My mistake. There are other commits that I do not want to merge.

PR #246, opened 2020-07-04
Add support for reading compressed MusicXML files

dmitrio95:
Currently Lomse is able to read uncompressed MusicXML files but refuses to open the compressed (`.mxl`) MusicXML files. This pull request implements a `Compiler` subclass for reading `.mxl` format which uncompresses the root MusicXML file and passes its content to the usual compiler for MusicXML format. The implementation of reading `.mxl` format follows [this reference](https://www.musicxml.com/tutorial/compressed-mxl-files/). Of course all this works only if Lomse is built with compression support, although I am not sure whether it makes sense to fully exclude `.mxl` compiler and the relevant methods from Lomse build in case compression support is disabled.

In general I tried to follow architecture and coding style seen around in the Lomse library but could certainly miss something. Also in some places I took a liberty to avoid manual creation and deletion of objects and arrays and used stack allocation and standard containers for managing their lifetime. This should cause no issues, at least performance ones, especially that C++11 standard with its move semantics is already [required](https://github.com/lenmus/lomse/blob/db7f2d410777020b2ab2ca379570afd5105062db/CMakeLists.txt#L431) by the build system. However if for some reasons this is undesirable for the project I can revert that part of the proposed changes.

### DIFF
--- a/add-sources.cmake
+++ b/add-sources.cmake
@@ -188,6 +188,7 @@ set(PARSER_FILES
 
     ${LOMSE_SRC_DIR}/parser/mxl/lomse_mxl_analyser.cpp
     ${LOMSE_SRC_DIR}/parser/mxl/lomse_mxl_compiler.cpp
+    ${LOMSE_SRC_DIR}/parser/mxl/lomse_compressed_mxl_compiler.cpp
 
     ${LOMSE_SRC_DIR}/parser/mnx/lomse_mnx_analyser.cpp
     ${LOMSE_SRC_DIR}/parser/mnx/lomse_mnx_compiler.cpp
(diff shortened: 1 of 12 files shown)


cecilios:
Thank you very much for this PR. I tested it and works perfectly!

> in some places I took a liberty to avoid manual creation and deletion of objects and arrays and used stack allocation and standard containers for managing their lifetime.

Good! Better use standard containers and avoid manual creation/deletion. My code is old and I've been learning modern C++ with it. There are place for lots of improvements.

> Of course all this works only if Lomse is built with compression support, although I am not sure whether it makes sense to fully exclude .mxl compiler and the relevant methods from Lomse build in case compression support is disabled.

Disabling compression was a request from a user wishing to minimize external dependencies (from zlib in this case) but accepting the limitation that it would not be possible to process, in future, compressed MusicXML files. The problem was not code size but not having to link against zlib. Probably very few people will build lomse without compression support but if anyone does, all code requiring to link against zlib must be excluded from the build. So, it is OK as it is.

dmitrio95:
OK, thanks for clarification!

PR #249, opened 2020-07-19
Delete measures jumps table and fix memory leaks

cecilios:
As pointed by @dmitrio95 ( https://github.com/lenmus/lomse/pull/248#pullrequestreview-449411287 ) the previous PR had a bug because the measures jumps table was never deleted. This PR fixes this. Also I've taken the opportunity to do a long debugging session for finding and fixing all memory leaks and other memory issues detected by Valgrind. Although most of the leaks where located in tests, some important ones were in the code.


cecilios:
@dmitrio95 Thanks for your feedback and comments. I learned C++ long time ago and I'm not yet very knowledgeable about the standard library and the new c++ features after c++11. So when I have doubts I prefer to write explicit code to ensure the objective. 

I have removed unnecessary invocation of std containers clear() method in destructors. Also fixed the std:unique_ptr initialization to nullptr. And re-checked with Valgrind.

PR #259, opened 2020-11-01
Enable compilation on Android

dmitrio95:
This pull request contains changes that are necessary to compile Lomse library for Android. Similar to #135, it was necessary to replace `ftello64`, `fseeko64` and `fopen64` functions which are not present on Android. In addition to that, I have added a change to `ioapi.h` which disables using 64-bit flavors of these functions when building for Android. This was necessary to get compressed files support in Lomse working on Android versions earlier than Android 7 (API level 24) since those versions didn't have 64-bit `fseeko`/`ftello` functions. This change disables support of compressed files larger than 4GB on that platform, but it doesn't seem to be needed for musical file formats anyway.

A similar change has also been done in [another fork](https://github.com/nmoinvaz/minizip/pull/138/files) of minizip, so synchronizing minizip code with that fork can also be another option to resolve compilation issues on some platforms.

### DIFF
--- a/packages/minizip/ioapi.c
+++ b/packages/minizip/ioapi.c
@@ -17,7 +17,7 @@
 #include "ioapi.h"
 
 //Lomse. Fix compilation on macOS, where ftello64, fseeko64 and fopen64 are not defined.
-#ifdef __APPLE__
+#if defined(__APPLE__) || defined(ANDROID)
     #define ftello64 ftello
     #define fseeko64 fseeko
     #define fopen64 fopen
--- a/packages/minizip/ioapi.h
+++ b/packages/minizip/ioapi.h
@@ -21,7 +21,8 @@
 #ifndef _ZLIBIOAPI64_H
 #define _ZLIBIOAPI64_H
 
-#if (!defined(_WIN32)) && (!defined(WIN32))
+// Lomse: excluded Android to avoid binary incompatibility with versions before Android 7
+#if (!defined(_WIN32)) && (!defined(WIN32)) && (!defined(ANDROID))
 
   // Linux needs this to support file operation on files larger then 4+GB
   // But might need better if/def to select just the platforms that needs them.


cecilios:
Hi Dmitri, thank you for the PR.

I do not think that the limitation to 4GB is important. Zip support is only needed for dealing with MusicXML files and, in future, with MNX files. And music files greater than 4GB probably do not exist, or if any, it must be an unusual exception, perhaps a file containing collections of music sheets. Thus, I do not see any problem with this PR.

As to your suggestion of basing Lomse in the minizip fork maintained by nmoinvaz, I didn't know about that fork. It is important to ensure that all third party code used in Lomse is properly maintained, as Lomse depends on it. The original minizip package, found in zlib distribution, has not been updated since October 2009. And it is clear that some maintenance is needed, at least to fix compilation warnings and some minor changes to deal with issues in other platforms, such as the ones addressed by your PR. Fork by nmoinvaz seems to be better maintained, so I wouldn't mind basing Lomse on it.

But the problem for doing this move is that the minizip package maintained by nmoinvaz has a totally different structure from the original minizip package. Probably nmoinvaz has done a very good job refactoring and updating the code but now, to use it it is necessary to determine what is needed for Lomse and doing some tests. Currently Lomse only uses two files from the minizip package (ioapi.c, unzip.c) and these files no longer exist in nmoinvaz fork. Thus, taking into account that:

a) minizip is very stable and it is rarely updated, and that Lomse only requires just unzip functionality, and
b) basing Lomse in minizip fork maintained by nmoinvaz requires some work to determine how to do it and do some testing, and do not have it now.

So I would vote to just apply your PR to current minizip code, and leave the move to nmoinvaz/minizip fork future, when more time available to do it or when forced by an important bug. Other ideas?

dmitrio95:
Yes, I agree that patching the current version of minizip would be fine. Moving to another fork may be a good option to make use of fixes that have already been added there, but if there are no serious issues with the old code I think it would be fine to stay with the current version of minizip.

cecilios:
Merged! Thank you.

PR #260, opened 2020-11-01
Make logger output destination customizable

dmitrio95:
This pull request is more intended for a discussion of the proposed changes, although it already implements a fully working solution at the moment.

The proposed change tries to enable applications to customize the way Lomse does its logging. Currently Lomse puts its logs to files at certain predefined paths which may be not suitable for all applications. An application may want to put logs to its application-specific directory, disable logging at all (like it has been done in https://github.com/hugbug/conpianist/issues/32) or redirect logs to another destination (standard output/error streams (`stdout`/`stderr`) or platform-specific logging systems like Android Logcat).

The proposal is to allow applications initialize logging system before creating `LomseDoorway` object by calling `lomse::get_global_logger().init()` and providing custom `std::ostream` objects to be used by the logger, for example:
```
lomse::get_global_logger().init(/* logStream */ &std::cout, /* forensicLogStream */ &std::cerr);
```
If custom log streams are not provided before creating `LomseDoorway` object, it initializes logging system to use the default logging configuration (the same predefined common and forensic log files as before).

Application can also deinitialize logging system in case the provided streams somehow get invalidated:
```
lomse::get_global_logger().deinit();
```

A somewhat similar approach is already used by Lomse allowing to provide `std::ostream` objects for reporting errors when calling `init_library()`, `new_document()` and `open_document()` methods in `LomseDoorway`, so this pull request expands this approach to cover also the library-global logging system.

Another possible approach could be allowing applications to provide custom `Logger` object with certain methods being overridden, which may potentially offer more flexibility in handling of logging severity levels.

Please let me know what do you think of this proposal, and whether another way of handling logging output would be more preferable.

### DIFF
--- a/include/lomse_logger.h
+++ b/include/lomse_logger.h
@@ -42,8 +42,6 @@ using namespace std;
 namespace lomse
 {
 
-extern ofstream dbgLogger;
-
 #if (LOMSE_COMPILER_MSVC == 1)
     #define __PRETTY_FUNCTION__       __FUNCTION__
 #endif
@@ -82,12 +80,23 @@ enum {
 class Logger
 {
 private:
+    std::ostream* m_logStream;
+    std::ostream* m_customForensicLogStream;
+    std::ofstream m_forensicLogStream;
     int m_mode;
     uint_least32_t m_areas;
+    bool m_initialized = false;
 
 public:
     Logger(int mode=k_normal_mode);
     ~Logger();
+
+    void init(std::ostream* logStream = nullptr, std::ostream* forensicLogStream = nullptr);
+    void deinit();
+    std::ostream& get_stream() { return *m_logStream; }
+
+    std::ostream& get_forensic_log_stream();
+    void close_forensic_log();
 
     inline bool debug_mode_enabled() { return m_mode == k_debug_mode; }
     inline bool trace_mode_enabled() { return m_mode == k_trace_mode; }
@@ -169,10 +178,34 @@ protected:
     void log_message(const string& file, int line, const string& prettyFunction,
                      const string& prefix, const string& msg);
     string format(const char* fmtstr, va_list args);
+
+    void clear_forensic_log();
+
+    std::string get_default_log_path();
 };
 
 extern Logger logger;
 
+inline Logger& get_global_logger() { return logger; }
+
+class StreamLogger
+{
+public:
+    template<typename T>
+    StreamLogger& operator<<(const T& t) {
+        logger.get_stream() << t;
+        return *this;
+    }
+
+    // Overload for I/O manipulators (e.g. std::endl)
+    StreamLogger& operator<<(std::ostream& (*manip)(std::ostream&)) {
+        logger.get_stream() << manip;
+        return *this;
+    }
+};
+
+extern StreamLogger dbgLogger;
+extern std::ofstream nullLogger;
 
 }   //namespace lomse
 
--- a/include/lomse_command.h
+++ b/include/lomse_command.h
@@ -205,7 +205,7 @@ protected:
     void log_forensic_data(Document* pDoc, DocCursor* pCursor);
     void set_command_name(const string& name, ImoObj* pImo);
     int validate_source(const string& source);
-    virtual void log_command(ofstream &logger);
+    virtual void log_command(ostream &logger);
 
 };
 
@@ -429,7 +429,7 @@ public:
 
 protected:
     void update_selection(SelectionSet* pSelection);
-    void log_command(ofstream &logger);
+    void log_command(ostream &logger);
 
 };
 
@@ -540,7 +540,7 @@ protected:
 
     //overrides and mandatory virtual methods
     void set_command_name();
-    void log_command(ofstream &logger);
+    void log_command(ostream &logger);
 
 };
 
@@ -598,7 +598,7 @@ public:
     ///@endcond
 
 protected:
-    void log_command(ofstream &logger);
+    void log_command(ostream &logger);
 };
 
 //---------------------------------------------------------------------------------------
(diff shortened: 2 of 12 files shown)


cecilios:
Hi Dmitri, thank you for the PR.

Currently I have no requirements for Lomse in this area but, as you point, there must be more flexibility for placing logs in other locations, redirecting them to other destinations or for disabling logging.

Your PR is certainly an improvement to allow for much more control of logging destinations. 

As to the suggested alternative of allowing applications to provide a custom Logger object with certain methods being overridden, my gut feeling is that this alternative could be more complex for Lomse users, being forced to derive from a Lomse base class. This could be good in certain scenarios but probably, the majority of Lomse users will not use it and could be more work for them to use Lomse. So I would vote for just applying your PR, that greatly improves the current situation. 

Really, I have no experience in solutions for application logging. Apart from the flexibility your PR provides, perhaps the most important requirement for me would be better facilities for dynamically enabling/disabling logging areas and more granularity on these areas. But the cost of the solutions I have devised for this is that logging code will be also included in release builds. So I have not found a good way to do it.

In conclusion, your PR addresses an important need, the code follows the library standards and looks OK to me. So, your PR is good for me. Nevertheless, before merging I will review your code more in detail and do some tests.

include/lomse_logger.h
```diff
@@ -169,10 +178,32 @@ class Logger
...
+
+    std::string get_default_log_path();
 };
 
 extern Logger logger;
 
+inline Logger& get_global_logger() { return logger; }
+
+class StreamLogger {
+public:
```
cecilios, line 192: Please curly brace between both lines.

include/lomse_logger.h
```diff
@@ -82,13 +80,24 @@ enum {
...
+    std::ofstream m_forensicLogStream;
     int m_mode;
     uint_least32_t m_areas;
+    bool m_initialized = false;
 
 public:
     Logger(int mode=k_normal_mode);
     ~Logger();
 
+    void init(std::ostream* logStream = nullptr, std::ostream* forensicLogStream = nullptr);
```
cecilios, line 94: In the proposed approach, if the applications would like to customize the logging system, it has to do it before creating the LomseDooorway object.

The drawback I see is that in my application and in the tutorials the LomseDoorway object is a member variable of the application class, and thus it is automatically created when the application class is created. In this scenario, initializing the logging system before creating the LomseDoorway object is not a trivial operation that requires changing some code. So, to simplify initialization and for backwards compatibility, the solution could be to change the constructor for the LomseDoorway, and adding to it the same parameters that the `Logger::init()` method:
```
LomseDoorway(std::ostream* logStream=nullptr, std::ostream* forensicLogStream=nullptr)
```
LomseDoorway will then initialize the logging system using those parameters. When nullptr is passed current logging destination will be used.

When `Logger.deinit()` is used the logging system can be re-initialized by using `Logger.init()`.

cecilios:
A question please. I'm obfuscated about how to disable logging! How would you do it?

dmitrio95:
> A question please. I'm obfuscated about how to disable logging! How would you do it?

Ah, sorry, I didn't mention that in my initial post.

With the currently implemented solution you cannot disable logging code completely (that would probably require more code changes) but you can still make logs be not printed anywhere by supplying an `std::ostream` which does nothing with its input. Probably the easiest way to do this is to pass an unopened `std::ofstream` instance:
```
static std::ofstream dummyStream;
lomse::get_global_logger().init(&dummyStream, &dummyStream);
```
The fix in https://github.com/hugbug/conpianist/issues/32 does effectively the same: all stream objects are still in place but they are never opened so everything that is being written to them just gets discarded.

This approach works for me as I just supply the needed streams depending on whether I need Lomse logs in this program run or not, but I see now that this is not really a user-friendly solution. Maybe it would be better to add also a separate method in `Logger` like `Logger::disable_logging()` which would internally do the same but wouldn't require users to figure out those tricks themselves?

On the second thought, there is also an already existing option to set logging areas to 0 for a `Logger`. That should already allow to disable most of logs, except that:
1) Log files will still be created;
2) Logging areas will not disable logging messages which use `dbgLogger` (or this PR's `logger.get_stream()`) directly;
3) Info/warn/error log messages don't currently seem to have logging areas assigned.

Maybe handling logging areas could also be expanded to avoid creating log files if all logging areas are disabled.

include/lomse_logger.h
dmitrio95: Done

include/lomse_logger.h
dmitrio95: Added these parameters to `LomseDoorway` constructor

cecilios:
For disabling logging it is necessary to avoid the same problem of having to initialize the logger before creating the LomseDoorway object. A simple solution would be to define the dummyStream in a lomse header file (better name it nullLogger?). In this way, disabling logging can be done also in LomseDoorway constructor:
```
LomseDorway lomse(&lomse::nullLogger, &lomse::nullLogger);
```

Please could you do it? Thank you.

when this PR is merged I will add documentation to the API docs to explain logging and how to initialize and use it.

> add also a separate method in Logger like `Logger::disable_logging()`

It would be nice, but I will not delay this PR for this. So add it only if you have time and would like to do it now.

> Maybe handling logging areas could also be expanded to avoid creating log files if all logging areas are disabled.

Well thought! Again a desirable improvement but only if you have time and would like to do it now.

dmitrio95:
Added `nullLogger` variable. I am not sure though whether its declaration should be added also to `lomse_doorway.h`: currently library users will need to include `lomse_logger.h` to use `nullLogger` but that still seems to be the most suitable place for that declaration.

As for the other improvements, I think it would be better to leave them for another PR: the possible logging areas improvements are mostly independent of the changes in this PR, and `nullLogger` should be a good way for disabling logs for now.

cecilios:
I think the place for the declaration of `nullLogger` should be lomse_logger.h but I was also hesitating about lomse_doorway.h

cecilios:
Thank you!

PR #265, opened 2020-11-07
Try to fix AppVeyor build

dmitrio95:
This PR tries to fix AppVeyor build which seems to be broken for a while in this repo.

The issue (possibly not the only one) seems to be that CMake build automatically tries to launch unit tests, but:
1. Path to the executable file seems to be determined incorrectly (this is something to be fixed separately though);
2. Correct libraries (at least `freetype.dll`) are not copied to the necessary location at that stage.

This makes build fail on `cmake --build .` execution stage, before AppVeyor build tries to copy the necessary libraries and launch the test executable.

This PR essentially restores `LOMSE_RUN_TESTS` option removed in cdc3a0fd171b58658ba2110f0b5b9bb5531d1f1b and turns it off in AppVeyor build to allow tests be launched as it is defined in `.appveyor.yml` configuration. I am not sure whether it helps to fully fix the build but at least chances are good that we get some new error messages to proceed with :)

If my guess is correct, 5ae011de16fbf69d1e3f4c9faeee6d4518e36853 might be the commit that triggered this build failure, since it enables `LOMSE_RUN_TESTS` option by default. However there might possibly be other relevant changes since then.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -43,7 +43,7 @@ build_script:
   # Compile Lomse
   - mkdir build
   - cd build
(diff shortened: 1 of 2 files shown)


cecilios:
Thank you!

If we restore the option to control whether to run tests or not, probably more changes are needed in CMakeLists.txt. But this change doesn't look to break anything. So, better let's fix the AppVeyor issue.

I have to recall why I removed that option. Probably it was in one of those days in which I had to build for Windows and was having problems. Windows is always a nightmare for me!

I will look with more detail to this in a few days.

cecilios:
I've been reviewing the changes to CMakeLists.txt and fully agree with restoring the LOMSE_RUN_TESTS option. Commit 0af607a3 fixes the documentation in CMakeLists.txt

PR #267, opened 2020-11-07
Fix a crash on reading MusicXML file with octave shift starting with a rest

### LINKED ISSUE #266: Crash on reading MusicXML file with octave shift starting with a rest
To reproduce the issue:
1. Try to load the attached MusicXML file ([test.musicxml.zip](https://github.com/lenmus/lomse/files/5504851/test.musicxml.zip)) with Lomse with the following code:
```
LomseDoorway lomse;
lomse.init_library(lomse::k_pix_format_rgba32, 96, false);
lomse.open_document(lomse::k_view_vertical_book, "test.musicxml");
```
Result: crash.

The attached file is a minimal example of score with octave shift starting with a rest:
изображение

If octave shift starts with a note, the file is loaded correctly.

I'll submit a (possible) fix for this issue soon.

dmitrio95:
Fixes #266.

This fix allows octave shifts start (and end) with `ImoNoteRest` rather than only `ImoNote`, and modifies the relevant places in the code (MusicXML reading and octave shifts handling in `StaffObjsCursor`) correspondingly. After these changes octave shifts starting with rests seem to be loaded and rendered correctly. However, I am not sure whether allowing octave shifts start only with notes was intentional, and whether I could miss something else with my changes.

### DIFF
--- a/include/lomse_mxl_analyser.h
+++ b/include/lomse_mxl_analyser.h
@@ -163,7 +163,7 @@ public:
     virtual ~MxlOctaveShiftBuilder() {}
 
     void add_relation_to_staffobjs(ImoOctaveShiftDto* pEndInfo) override;
-    void add_to_open_octave_shifts(ImoNote* pNote);
+    void add_to_open_octave_shifts(ImoNoteRest* pNR);
 };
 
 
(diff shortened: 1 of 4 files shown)


cecilios:
Thanks for the PR Dmitri,

> I am not sure whether allowing octave shifts start only with notes was intentional, and whether I could miss something else with my changes.

I didn't included rests because I though than an octave change will always start and end in a note, not in a rest. And all MusicXML samples I had with octave changes seemed to confirm that guess. Your PR looks OK to me.

I'm going now to build and check your PR to confirm that there are no issues in regression tests, before merging.

cecilios:
All regression tests passed! Thank you.

PR #270, opened 2020-11-29
Changes to support Mac OSX

grantsloman:
A few changes for building & running on Mac OSX. WxWidgets tutorials 1 & 2 now build and run successfully. Further work is required to get wxMidi working for tutorials 3 & 4.

Also a few general Unix fixes for copying target libs & creating symbolic link.


include/lomse_im_note.h
```diff
@@ -134,9 +134,9 @@ class ImoNoteRest : public ImoStaffObj
     ImoTuplet* get_first_tuplet();
 
     //IM attributes interface
-    virtual void set_int_attribute(TIntAttribute attrib, int value);
-    virtual int get_int_attribute(TIntAttribute attrib);
-    virtual list<TIntAttribute> get_supported_attributes();
+    virtual void set_int_attribute(TIntAttribute attrib, int value) override;
```
cecilios, line 137: Thanks for this PR. 
The changes in this file create a conflict. Probably your fork is not updated to latest code. The 'override' atrribute was fixed in all lomse code  (I hope) in a previous PR and, for sure, it is already fixed in this file. Please remove this file from the PR. Thank you

build-version.cmake
```diff
@@ -28,7 +28,7 @@ if (EXISTS ${LOMSE_ROOT_DIR}/.git)
       WORKING_DIRECTORY "${LOMSE_ROOT_DIR}"
       OUTPUT_VARIABLE LOMSE_VERSION_BUILD
       OUTPUT_STRIP_TRAILING_WHITESPACE)
-    execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --long --dirty
+    execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --long --dirty --all
```
cecilios, line 31: Not sure if this is correct. I have to test and investigate. On my local branch, when I execute `git describe --tags --long --dirty` I get `0.27.0-114-g965c2f91` but when I execute  `git describe --tags --long --dirty --all` the result is `heads/changes-0-g965c2f91` and the version is lost! Why did you need to change this?

cecilios (changes requested):

CMakeLists.txt
```diff
@@ -197,6 +197,9 @@ cmake_policy(SET CMP0042 NEW)   #MACOSX_RPATH is enabled by default
 #Windows 10 camplains 
 cmake_policy(SET CMP0054 NEW)   # quoted arguments will not be further dereferenced
 
+#OSX complains 
+cmake_policy(SET CMP0087 NEW)   # evaluate generator expressions for install(CODE) and install(SCRIPT)
```
cecilios, line 201: Policy CMP0078 was defined in CMake version 3.13. As the minimum required version for building lomse is 2.8.10, any version lower than 3.13 generates an error:
```
CMake Error at CMakeLists.txt:208 (cmake_policy):
  Policy "CMP0087" is not known to this version of CMake.
```
Please change to:

```
if (${CMAKE_VERSION} GREATER "3.12")
    cmake_policy(SET CMP0078 NEW)
endif()
```

cecilios (changes requested):

build-version.cmake
grantsloman: It was an attempt to work around the fatal error as shown below...
```
Grants-iMac:lomse grant$ git describe --tags --long --dirty
fatal: No names found, cannot describe anything.
Grants-iMac:lomse grant$ git describe --tags --long --dirty --all
heads/master-0-gc576c38
```

CMakeLists.txt
grantsloman: Done

include/lomse_im_note.h
grantsloman: Reverted (I hope!)

build-version.cmake
grantsloman: I've reverted this change for now

CMakeLists.txt
grantsloman: Done

build-version.cmake
cecilios: Thank you. Now I understand the problem. I assumed that the repo always contains the tags and thus, that it was feasible to extract the version from the latest tag. But as this link explains (https://stackoverflow.com/questions/4916492/git-describe-fails-with-fatal-no-names-found-cannot-describe-anything) current solution doesn't work for shallow clones.

By adding `--all` you avoid the error for shallow clones, but prevents also that it works for clones with tags, making that code useless. So adding `--all` is not the solution. But it is necessary to avoid the 'fatal: No names found,..." message as not only it is not fatal but also it is irrelevant! I will investigate how to do it. So for now better leave it without the `--all` as you already did. Thank you.

CMakeLists.txt
dmitrio95: It seems you have added `CMP0087` first and later changed it to `CMP0078`?

Anyway, if your changes require the new behavior for this policy (namely, generator expressions support for `install`), wouldn't this break things for earlier versions of CMake which don't have support for this feature included?

grantsloman: In my change I've kept it as CMP0087 - expecting that 78 was a typo intended to be 87.

If I remember correctly, I only added that to prevent a warning so might not be critical.

cecilios: @dmitrio95 Thanks for spotting this! My fault. I swapped the numbers in my comment !

I agree that the NEW behavior should only be enforced when it does not break things. As usual, CMake documentation is rubbish: no explanation of the problem that this new behavior solves, no examples, no tips for changes to do in cmake scripts, ... just template garbage to fill four paragraphs. And by googling for CMP0087 I did not find  anything relevant.

My conclusion was that generator expressions (as defined here https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html ) are now accepted and evaluated in `install()` sentences. But, as lomse cmake script does not use generator expressions in install commands, it is irrelevant what behavior is used. Thus, as the old behavior is deprecated, it would be better to enable the new behavior to have the script better prepared in case of future changes in `install()` statements. And in any case, to test the PR before merging, to check that nothing is broken.

Any thoughts?

dmitrio95: > In my change I've kept it as CMP0087 - expecting that 78 was a typo intended to be 87.

The number is correct in the commit message but still seems to be reverted in the change itself: https://github.com/lenmus/lomse/pull/270/commits/5f7cb5c26524e36c1ee8e796dda9a27a76f84cba

> as lomse cmake script does not use generator expressions in install commands, it is irrelevant what behavior is used

The warning is probably a result of the change made it this PR which does add a generator expression to an `install` command (see [these lines](https://github.com/lenmus/lomse/pull/270/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR887-R890)). If I understand the documentation [here](https://cmake.org/cmake/help/latest/policy/CMP0087.html) correctly, this feature is only supported since CMake 3.14 so earlier versions of CMake may be unable to handle this correctly. That's why I was asking about this: the change will probably work with newer versions of CMake but may lead to unexpected results with versions before 3.14.

grantsloman: My apologies, you're absolutely right. I had fully intended it to be CMP0087 but seems that I've accidentally copy/pasted it as 78 from Cecilio's comment above.

Creating a symbolic link to the .so wasn't working, hence why I changed it to 
```
ln -fs ${CMAKE_INSTALL_PREFIX}/lib/$<TARGET_FILE_NAME:lomse-shared> 
                                        ${CMAKE_INSTALL_PREFIX}/lib/liblomse.so
```
which then worked correctly.
Really sorry but I'll have to leave this one with you to figure out - or revert the change, if that's necessary (although it took me a long time to figure out why linking to the shared library was failing!)

cecilios: @dmitrio95 You are right! I overlooked that a generator expression was introduced. I'm going to test this PR with cmake < 3.14 and also investigate the issue of the symlink commented by @grantsloman

cecilios: PR tested. When using cmake version 3.10.2, as expected, the generator expression is not evaluated and the resulting command is
```
ln -fs /usr/local/lib/$<TARGET_FILE_NAME:lomse-shared> /usr/lib/liblomse.so
```
and fails to create the symlink. So it is clear that CMP0087 NEW behavior is not acceptable unless forcing to require a cmake >= 3.14

@grantsloman To try understand why the unpatched `CMakeLists.txt` failed to create the symlink in macOS it would be useful to look at the generated command. And this is simple because in build folder cmake generates a file named `"cmake_install.cmake"`. Please open it with a text editor and look for string `"EXECUTE_PROCESS(COMMAND ln -fs"`. There is only one occurrence, around line 78, e.g. (when using cmake 3.10.2):
```
  EXECUTE_PROCESS(COMMAND ln -fs /usr/local/lib/$<TARGET_FILE_NAME:lomse-shared> 
                                        /usr/local/lib/liblomse.so )
```
There you will see the `ln` command and its arguments and you can try to find why it does not work for you.

cecilios: @grantsloman
A posible solution to supress `fatal: No names found, cannot describe anything. ` message is to add `ERROR_QUIET` parameter to `execute_command`:
```
    execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --long --dirty
      WORKING_DIRECTORY "${LOMSE_ROOT_DIR}"
      OUTPUT_VARIABLE LOMSE_VERSION_GIT
      OUTPUT_STRIP_TRAILING_WHITESPACE
      ERROR_QUIET)
```
I have tested and it works.

grantsloman: If I change the CMakeLists.txt line back to
```
"EXECUTE_PROCESS(COMMAND ln -fs ${CMAKE_INSTALL_PREFIX}/lib/${LOMSE_SHARED} 
                                        ${CMAKE_INSTALL_PREFIX}/lib/liblomse.so )" 
```
then the resulting generated command in `cmake_install.cmake` is
```
EXECUTE_PROCESS(COMMAND ln -fs /usr/local/lib/lomse-shared 
                                        /usr/local/lib/liblomse.so )
```
which isn't the correct target file name.

cecilios: I've done some tests and, curiously, the symlink is always created even when removing these lines:
```
    #create symlink to library
	install( CODE
        "EXECUTE_PROCESS(COMMAND ln -fs ${CMAKE_INSTALL_PREFIX}/lib/${LOMSE_SHARED} 
                                        ${CMAKE_INSTALL_PREFIX}/lib/liblomse.so )" 
    )
```

Therefore it seems irrelevant if that command is correct or not, and can be removed.

It looks as if cmake automatically creates the library and the symlink at build time. You can find them in build folder. Please notice that the previous command does not create a symlink in build folder so it is automatically created by cmake. And both are copied to `/usr/local/lib` at install time, and the symlink is adjusted to point to the installed library.

I have tested removing the previous lines from CMakeLists.txt and it works perfectly and the symlink is created at installation time.

I'm not been able to find any cmake documentation explaining this and confirming that symlinks are automatically created for libraries, but my tests suggest this. I am missing something?

@grantsloman Could you please test this, removing previous lines, in macOS?

PR #272, opened 2020-12-06
Correct engraving of volta located at the start of a system

dmitrio95:
This pull request corrects appearance of volta which happen to start at the first measure of a system.

The issue can be reproduced, for example, on a score from [`test-scores/unit-tests/repeats/04-repeat-barline-simple-volta.xml`](https://github.com/lenmus/lomse/blob/1f7b2235e8690f6a4c3de4cfe1cf0ff27d3a3373/test-scores/unit-tests/repeats/04-repeat-barline-simple-volta.xml). If you open this file with Lomse and set page width so that the second volta starts on a new system, the score is rendered as follows (note that number `2` and some part of a bracket are drawn after the end of the first system):
before

The reason of this issue seems to be that volta brackets are bound to barlines in the internal model, and the relevant start barline for a volta is actually located at the end of the previous system. Volta engraver doesn't account for that though and starts rendering volta bracket from that point. The proposed commit avoids creating a volta bracket shape near the start barline in this case so volta brackets get rendered as expected:
after

The fact that the start barline is situated at system end is currently detected just by doing a (fuzzy) comparison of the corresponding shape coordinates with the right boundary of the current system. This works (at least in those cases I tested) but doesn't seem very robust in general, so please let me know if there is a better way to detect whether a particular barline is the last in the current system.

### DIFF
--- a/include/lomse_volta_engraver.h
+++ b/include/lomse_volta_engraver.h
@@ -51,6 +51,7 @@ class VoltaBracketEngraver : public RelObjEngraver
 {
 protected:
     int m_numShapes;
+    bool m_fFirstShapeAtSystemStart = false;
     ImoVoltaBracket* m_pVolta;
     LUnits m_uStaffTop;             //top line of current staff
     LUnits m_uStaffLeft;
--- a/src/graphic_model/engravers/lomse_volta_engraver.cpp
+++ b/src/graphic_model/engravers/lomse_volta_engraver.cpp
@@ -165,6 +165,13 @@ GmoShape* VoltaBracketEngraver::create_first_shape()
 {
     //first shape when there are more than one
 
+    if (!m_fFirstShapeAtSystemStart && m_pStartBarlineShape->get_x_right_line() - m_uStaffRight > -1.0f)
+    {
+        //start barline is at the end of a system, create first shape at next system start instead
+        m_fFirstShapeAtSystemStart = true;
+        return nullptr;
+    }
+
     GmoShapeVoltaBracket* pShape = LOMSE_NEW GmoShapeVoltaBracket(m_pVolta, m_numShapes, m_color);
     pShape->enable_final_jog(false);
 
@@ -212,23 +219,14 @@ void VoltaBracketEngraver::set_shape_details(GmoShapeVoltaBracket* pShape,
     LUnits xStart = m_uStaffLeft;
     LUnits xEnd = m_uStaffRight;
 
-    if (shapeType == k_single_shape)
+    if (!m_fFirstShapeAtSystemStart && (shapeType == k_single_shape || shapeType == k_first_shape))
     {
         xStart = m_pStartBarlineShape->get_x_right_line();
         if (m_pStartBarlineShape->get_width() < 40.0f)
             xStart += 30.0f;
+    }
 
-        xEnd = m_pStopBarlineShape->get_x_left_line();
-        if (m_pStopBarlineShape->get_width() < 40.0f)
-            xEnd -= 30.0f;
-    }
-    else if (shapeType == k_first_shape)
-    {
-        xStart = m_pStartBarlineShape->get_x_right_line();
-        if (m_pStartBarlineShape->get_width() < 40.0f)
-            xStart += 30.0f;
-    }
-    else if (shapeType == k_final_shape)
+    if (shapeType == k_single_shape || shapeType == k_final_shape)
     {
         xEnd = m_pStopBarlineShape->get_x_left_line();
         if (m_pStopBarlineShape->get_width() < 40.0f)


cecilios:
Thank you for spotting this and sending this PR. I've reviewed it, and it works perfectly and no issues with regression tests. So, I'm in favor of merging it.

About the question for a better way to detect whether a particular barline is the last in the current system, currently the fact that a barline is the last one in the system is not recorded and, thus, not passed to engravers. The test in your PR:
```
if (!m_fFirstShapeAtSystemStart && m_pStartBarlineShape->get_x_right_line() - m_uStaffRight > -1.0f)
```
works well in all my tests. Replacing it by `abs(m_pStartBarlineShape->get_x_right_line() - m_uStaffRight) < 1.0f` does not work in all tests (!).

The `!m_fFirstShapeAtSystemStart && ` part is not necessary but doesn't harm. 

During the tests, I've found another bug in Lomse code: method `create_intermediate_shape()` is never invoked. Test score `test-scores/50201-repeat-barlines-split-volta.xml` is rendered as follows:
image

If window width is progressively reduced we arrive at this:

image

As you can see, the 2nd bracket should generate a line on second staff. I've located the origin of the problem but solving it will require some changes in the engraving process, and I have to think how to do it. I will also take the opportunity to add a flag to inform the engraver that the barline is the last one in the system.

If you agree, I will merge your PR.

Just for your information, currently, the engraving process is as follows. It is the standard way to engrave all RelObjs; here the only difference is that there are no intermediate barlines as only start and end barlines are added to the relationship. 

### When both barlines, star and end, are in the same system:###

When processing start barline, method ``ShapesCreator::start_engraving_relobj()`` is invoked. It:
- creates the engraver
- invokes its ``set_start_staffobj()`` method
- and saves the engraver, associated to the RelObj (the ImoVoltaBracket object)

Finally, when processing the end barline, method ``ShapesCreator::finish_engraving_relobj()`` is invoked. It:
- retrieves the engraver
- invokes its ``set_end_staffobj()`` method
-  And then method ``create_last_shape()`` is invoked


### When start barline is in a different system than end barline the process is as follows:###

When processing start barline, as before, method ``ShapesCreator::start_engraving_relobj()`` is invoked. It:
- creates the engraver
- invokes its ``set_start_staffobj()`` method
- and saves the engraver, associated to the RelObj

When processing the end barline, the SystemLayouter detects that it is in a different system than start barline and, thus, it invokes ``engrave_not_finished_relobj()`` to finish engraving first system. Then it:
- invokes ``ShapesCreator::create_first_or_intermediate_shape()`` and it:
    - retrieves the engraver
    - invokes its ``create_first_or_intermediate_shape()`` method

Then SystemLayouter continues engraving all RelObjs associated to the end barline and, thus:
- retrieves the VoltaBracket engraver
    - invokes its ``set_end_staffobj()`` method
    - And then method ``create_last_shape()`` is invoked

dmitrio95:
> The `!m_fFirstShapeAtSystemStart && ` part is not necessary but doesn't harm.
> 
> During the tests, I've found another bug in Lomse code: method `create_intermediate_shape()` is never invoked.

Yes, I have also noticed the issue with `create_intermediate_shape()` for wedges and octave shifts, so this seems to be the case at least for some of RelObjs.

However, if I understand the idea of engravers system correctly, the `!m_fFirstShapeAtSystemStart` check should be necessary for voltas spanning for more than one system.

As far as I can tell, the idea is that engraver of `RelObj` is expected to receive consecutive `create_first_or_intermediate_shape()` calls for each system except the last one and a `create_last_shape()` call for its last system. Therefore if a volta has its start barline on system 1, spans for the entire system 2 and ends on system 3, `create_first_or_intermediate_shape()` would be called two times. Volta engraver (as some other engravers) redirects these calls to `create_first_shape()` if it hasn't create any shapes so far. If start barline is located at the end of system 1 (the case covered by this PR), we would want to skip the first call of `create_first_shape()` but not the second one (as we actually want the first volta bracket shape to cover system 2). The `!m_fFirstShapeAtSystemStart` part of the check is intended to do exactly this: we should do the check during the first call of `create_first_shape()`, but the second call should not be skipped and should return a valid shape.

> As you can see, the 2nd bracket should generate a line on second staff. I've located the origin of the problem but solving it will require some changes in the engraving process, and I have to think how to do it. I will also take the opportunity to add a flag to inform the engraver that the barline is the last one in the system.
> 
> If you agree, I will merge your PR.

Yes, if there is currently no better way to retrieve this barline position information, I think it would be fine to apply the current solution and change it later if such way appears.

> Just for your information, currently, the engraving process is as follows

Thanks for the information! I have so far tried adding pedal support to Lomse which gave me some very basic impression on how engraving of such objects works here, but I didn't dig much into details yet.

While at it, I would also like to take an opportunity to ask a question on this: so far I got pedal marks and lines rendered in most basic cases (which I did just similarly to the way it is done for wedges and octave shifts) but I haven't yet figured out how it would be more natural to support "intermediate" points like pedal changes (`_/\_`-like marks) and unpaired pedal down/pedal up signs (not sure if they are allowed by MusicXML but it looks like they might). I see at least two options to do this:

1. Add all the relevant `ImoDirection`s or notes to the pedal relation.
2. Add separate entities for pedal marks and lines (`ImoPedalMark` objects and `ImoPedalLine` relations between them).

The first solution is simpler but seems to make it more difficult to handle unpaired pedal marks (unless pedal relations get allowed to start and end on the same object?) The second one, on the other side, would introduce more different object types to deal with, but perhaps it would also be more correct in the context of Lomse internal model design? If you could give some hints or examples on how such cases are better to be handled in Lomse that would be very helpful.

cecilios:
> As far as I can tell, the idea is that engraver of RelObj is expected to receive consecutive create_first_or_intermediate_shape() calls for each system except the last one and a create_last_shape() ....

You are absolutely right! The `!m_fFirstShapeAtSystemStart &&` must be part of the check.


> While at it, I would also like to take an opportunity to ask a question on this:

Feel always free to as questions when needed, I'll try to help. Use PM if preferred or more appropriate.

I will merge now your PR and then to study your questions. I will try to answer this as soon as possible, probably this evening or tomorrow.

Thanks for contributing!

cecilios:
The answer to your questions. 

After many years dealing with music notation I've learned that it is so complex than it is better to be prepared for unexpected requirements. Thus, when having doubts about two alternatives, I usually opt for the most flexible, unless much more higher implementation costs for one of them or when there are high computational costs for managing one of them.

When having to add a new feature, I usually start by studying the MusicXML model, as it is complex enough and Lomse has to support it. And normally, that is enough to have a clear picture and take the initial decisions.

After studying the MusicXML approach for pedal marks, I understand that:
- pedal marks with line="yes" are a kind of spanner direction, similar to octave changes and wedges, but can have intermediate points.
- pedal marks with line="no" represent just a symbol and type="start, stop" determines the symbol to use. They are single symbol `<direction>` similar to other such as coda, segno and dynamics
- The intermediate pedal marks, indicated with an inverted V like marking, are represented by a pedal element with line="yes" and type="change"

Thus we have two types of pedal marks: single symbol, and spanner lines. Therefore, I am in favor of creating two objects:
- The `ImoPedalMark`, a direction that models a symbol. For flexibility and in line with other single symbol `<direction>` elements such as coda, segno and dynamics, `ImoPedalMark` would be an `ImoAuxObj` attached to an `ImoDirection`.
- The `ImoPedalLine` would be an `ImoRelObj` relating two or more `ImoDirection` objects having an `ImoPedalMark` attachment.

Some MusicXML `<pedal>` directions will create just one `ImoPedalMark` or an `ImoPedalLine`. Others will create both.

I tentatively would start by creating the `ImoPedalMark` (`ImoAuxObj`) and `ImoPedalLine` (`ImoRelObj`) objects and modifying the MusicXML analyser to handle pedal marks, to get a better feeling. I would use `WedgeMxlAnalyser` as a starting point. If nothing found to change the decision, then I would proceed with the engravers.

Having two pedal mark related objects implies to have two engravers, one for `ImoPedalMark` objects and another for `ImoPedalLine` objects.

The `ImoPedalLine` engraver has to engrave only lines and looks simple. I would use `OctaveShiftEngraver` as a starting point.

The engraver for `ImoPedalMark` objects looks also simple. I would use `CodaSegnoEngraver` as a starting point.


Hope this helps. If you have doubts or more questions, please feel free to ask.

dmitrio95:
Thanks for your answer! I'll try implementing the solution with two objects then. My current incomplete solution handles mostly the case of pedal lines so I'll split handling of pedal marks off that and try to get all this working together.

PR #277, opened 2020-12-12
Correct engraving of wedge when at end of system

cecilios:
This PR fixes a similar issue to #272, but for wedges, when the wedge starts just before the last barline in a system. See, for instance, the two small segments at end of first system in this image (tests-scores/dev-tests/spanners/wedge-02-start-at-end-of-system.xml):
image

I have also checked in octave-shifts but there the issue is not present there because the octave-shift will be attached to a note. See, for instance, tests-scores/dev-tests/spanners/octave-shift-02-start-at-end-of-system.xml.

I have fixed the issue in wedges following the same approach that you did in volta brackets.

As to how to detect that we are at end of a system, in the case of a wedge it is a ImoDirection preceding a ImoBarline and, thus, detecting that the barline is the last on in the system is of no help. The solution has been to require at least a certain amount of space to start a wedge. I have requested at least 2 staff lines space. And I have applied the same approach to volta brackets, so now I believe that the check is much stronger and no need to detect that the barline is the last one in the system. Any thoughts?

### DIFF
--- a/src/graphic_model/engravers/lomse_volta_engraver.cpp
+++ b/src/graphic_model/engravers/lomse_volta_engraver.cpp
@@ -174,7 +174,9 @@ GmoShape* VoltaBracketEngraver::create_first_shape()
 {
     //first shape when there are more than one
 
-    if (!m_fFirstShapeAtSystemStart && m_pStartBarlineShape->get_x_right_line() - m_uStaffRight > -1.0f)
+    LUnits minLength = tenths_to_logical(10.0f);
+    if (!m_fFirstShapeAtSystemStart
+        && abs(m_uStaffRight - m_pStartBarlineShape->get_x_right_line()) < minLength)
     {
         //start barline is at the end of a system, create first shape at next system start instead
         m_fFirstShapeAtSystemStart = true;
--- a/src/graphic_model/engravers/lomse_volta_engraver.cpp
+++ b/src/graphic_model/engravers/lomse_volta_engraver.cpp
@@ -176,7 +176,7 @@ GmoShape* VoltaBracketEngraver::create_first_shape()
 
     LUnits minLength = tenths_to_logical(10.0f);
     if (!m_fFirstShapeAtSystemStart
-        && abs(m_uStaffRight - m_pStartBarlineShape->get_x_right_line()) < minLength)
+        && m_uStaffRight - m_pStartBarlineShape->get_x_right_line() < minLength)
     {
         //start barline is at the end of a system, create first shape at next system start instead
         m_fFirstShapeAtSystemStart = true;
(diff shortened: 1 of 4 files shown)


src/graphic_model/engravers/lomse_volta_engraver.cpp
```diff
@@ -174,7 +174,9 @@ GmoShape* VoltaBracketEngraver::create_first_shape()
 {
     //first shape when there are more than one
 
-    if (!m_fFirstShapeAtSystemStart && m_pStartBarlineShape->get_x_right_line() - m_uStaffRight > -1.0f)
+    LUnits minLength = tenths_to_logical(10.0f);
+    if (!m_fFirstShapeAtSystemStart
+        && abs(m_uStaffRight - m_pStartBarlineShape->get_x_right_line()) < minLength)
```
dmitrio95, line 179: I agree with the general approach to require a minimal amount of space here, but do we need to take an `abs` of the difference in this check? The same change in `WedgeEngraver` doesn't contain it, and the overall condition seems to be more correct if not taking an absolute value from the coordinates' difference.

cecilios: In theory the last barline shape right border should be aligned with the staff right border, and thus the difference should be always zero. I've looked at some dumps of the graphic model and this was the case. But we are dealing with floating point numbers, many computations, truncation errors and really low space units (one cent of a millimeter). I'm not 100% sure that the barlines left border will always coincides exactly with the staff right border, and the error could be positive or negative. So the `abs` is to ensure than any small floating point difference does not cause problems here. I'm not comfortable removing it. Why do you think it is not necessary?

But in the wedge engraver, the wedge is attached to an ImoDirection. And in the worst case it will be placed just before the barline. Thus, the difference between ImoDirection shape (a GmoInvisibleShape) an the staff right border will be always a positive number (the barline shape width). Now, thinking on this, if the barline where not visible (MusicXML attribute `print-object="no"`), the barline width will be zero and thus, it is also necessary to put the `abs`!

dmitrio95: To make sure we are talking about the same thing, I was suggesting to rewrite the second part of the condition as
```
m_uStaffRight - m_pStartBarlineShape->get_x_right_line() < minLength
```
so we don't take `abs` of the left part of this condition.
In this form it will work perfectly regardless of whether errors are positive or negative: as `minLength` is a positive number, any negative number will be less than it. So with this condition the first shape would be created at the current system if the barline's right edge is located at least `minLength` to the left from the staff's right edge, while the condition with `abs` also allows that barline be located at `minLength` to the *right* from the staff's edge. The latter situation should not happen anyway so both forms would effectively be equivalent in this code. However that also means that `abs` can be freely removed from this condition, and the form without `abs` seems to me to more correctly describe the meaning of this check. That is why I suggested removing `abs`, but both forms would be fine here.

cecilios: Thank you for your comments. I didn't pay attention than any negative number will be less than `minLength`. You are absolutely right and `abs` is not necessary!.

PR #279, opened 2020-12-16
Attach dynamics markings to notes or rests on reading MusicXML files

dmitrio95:
Currently dynamics markings engraver assumes that dynamics is attached to a note or rest. MusicXML importer however attaches it to `ImoDirection` when reading it as a descendant of `<direction>` tag. This difference results in dynamics being not aligned with notes or rests which they belongs to, for example (with [`test-scores/50035-directions-take-no-space.xml`](https://github.com/lenmus/lomse/blob/fe7e7bc993c5b83a5fee5b8bf682d5ead21b6d9f/test-scores/50035-directions-take-no-space.xml))
before

With some real-world scores it may look even worse: some dynamics marks become crossing barlines and causing some wedges being pushed further down:
Screenshot_20201216_092728

This commit makes MusicXML analyser attach dynamics markings to notes and rests instead of `ImoDirection`. I am not sure whether it has any implications on the model side, but the engraver seems to work much better in this case making dynamics marks aligned with the relevant notes:
after

Screenshot_20201216_092919

The approach used here is similar to that already used for octave shifts: if dynamics marking needs to be attached to a note or rest which is not yet available, it is added to a list of pending items which get attached to the next note or rest when it gets read. This approach should be possible to use for other objects that might need it as well — at least I was considering something similar for pedal marks (unless they would be better to be attached to `ImoDirection`?).

### DIFF
--- a/src/parser/mxl/lomse_mxl_analyser.cpp
+++ b/src/parser/mxl/lomse_mxl_analyser.cpp
@@ -2806,6 +2806,10 @@ public:
         error_if_more_elements();
 
         pSO->add_attachment(pDoc, pImo);
+
+        if (!pSO->is_note_rest())
+            m_pAnalyser->add_pending_dynamics_mark(pImo);
+
         return pImo;
     }
 
@@ -4233,6 +4237,7 @@ public:
 
         pNR->set_visible(fVisible);
         add_to_model(pNR);
+        attach_pending_dynamics_marks(pNR);
         add_to_spanners(pNR);
 
         //deal with grace notes
@@ -4568,6 +4573,12 @@ protected:
             m_pBeamInfo->set_note_rest(pNR);
             m_pAnalyser->add_relation_info(m_pBeamInfo);
         }
+    }
+
+    //----------------------------------------------------------------------------------
+    void attach_pending_dynamics_marks(ImoNoteRest* pNR)
+    {
+        m_pAnalyser->attach_pending_dynamics_marks(pNR);
     }
 
     //----------------------------------------------------------------------------------
@@ -7741,6 +7752,23 @@ void MxlAnalyser::clear_pending_relations()
 
     m_lyrics.clear();
     m_lyricIndex.clear();
+    m_pendingDynamicsMarks.clear();
+}
+
+//---------------------------------------------------------------------------------------
+void MxlAnalyser::attach_pending_dynamics_marks(ImoNoteRest* pNR)
+{
+    for (ImoDynamicsMark* pDynamics : m_pendingDynamicsMarks)
+    {
+        ImoContentObj* pOldParent = pDynamics->get_contentobj_parent();
+
+        if (pOldParent)
+            pOldParent->remove_but_not_delete_attachment(pDynamics);
+
+        pNR->add_attachment(m_pDoc, pDynamics);
+    }
+
+    m_pendingDynamicsMarks.clear();
 }
 
 //---------------------------------------------------------------------------------------
--- a/include/lomse_mxl_analyser.h
+++ b/include/lomse_mxl_analyser.h
@@ -250,6 +250,7 @@ protected:
     MxlVoltasBuilder*   m_pVoltasBuilder;
     MxlWedgesBuilder*   m_pWedgesBuilder;
     MxlOctaveShiftBuilder*  m_pOctaveShiftBuilder;
+    vector<ImoDynamicsMark*> m_pendingDynamicsMarks;
     map<string, int>    m_lyricIndex;
     vector<ImoLyric*>   m_lyrics;
     map<string, int>    m_soundIdToIdx;     //conversion sound-instrument id to index
@@ -407,6 +408,10 @@ public:
 
     //interface for building beams
     inline bool fix_beams() { return m_libraryScope.get_musicxml_options()->fix_beams(); }
+
+    //interface for building dynamics marks
+    void add_pending_dynamics_mark(ImoDynamicsMark* pObj) { m_pendingDynamicsMarks.push_back(pObj); }
+    void attach_pending_dynamics_marks(ImoNoteRest* pNR);
 
     //interface for building lyric lines
     void add_lyrics_data(ImoNote* pNote, ImoLyric* pData);
--- a/include/private/lomse_internal_model_p.h
+++ b/include/private/lomse_internal_model_p.h
@@ -2356,6 +2356,7 @@ public:
     ImoAuxObj* get_attachment(int i);
     void add_attachment(Document* pDoc, ImoAuxObj* pAO);
     void remove_attachment(ImoAuxObj* pAO);
+    void remove_but_not_delete_attachment(ImoAuxObj* pAO);
     ImoAuxObj* find_attachment(int type);
 
     //overrides for Observable children
--- a/src/internal_model/lomse_internal_model.cpp
+++ b/src/internal_model/lomse_internal_model.cpp
@@ -2154,6 +2154,13 @@ void ImoContentObj::remove_attachment(ImoAuxObj* pAO)
 }
 
 //---------------------------------------------------------------------------------------
+void ImoContentObj::remove_but_not_delete_attachment(ImoAuxObj* pAO)
+{
+    ImoAttachments* pAuxObjs = get_attachments();
+    pAuxObjs->remove(pAO);
+}
+
+//---------------------------------------------------------------------------------------
 ImoAuxObj* ImoContentObj::find_attachment(int type)
 {
     ImoAttachments* pAuxObjs = get_attachments();


cecilios:
Thanks for this PR.

Lomse philosophy is that any StaffObj can have attachments (AuxObjs and RelObjs). AuxObjs are isolated notations and the only restrictions about the parent staff obj are the ones we would like to impose due to semantic reasons or other, but not due to the internal model. In the case of dynamic marks there is no problem in attaching them to notes/rests. In fact, in the LDP language format, dynamic-marks are attached to notes/rests and this is the reason for considering also note heads in the engraver.

I have doubts about dynamic marks in MusicXML because the spec. says that "Dynamics can be associated either with a note or a general musical direction". I have to study this in more detail before commenting your PR. I will do it ASAP.

cecilios:
According to the MusicXML spec. _"Dynamics can be associated either with a note or a general musical direction"_. After reviewing the spec. with more detail I understand than dynamics marks (`<dynamics>` elements) can be child of `<notations>` and of `<direction>` elements. In the first case, Lomse will attached it to the note having the `<notations>` child. In the second case the current Lomse approach is to attach it to an `ImoDirection` object. In the lilypond samples (folder `tests-scores/regression/scores/lilypond`) there are examples of both cases, e.g. `43e-Multistaff-ClefDynamics.xml` line 42 (attached to direction) and `42b-MultiVoice-MidMeasureClefChange.xml` line 255 (attached to note).

The MusicXML spec. also says that _"For applications where a specific direction is indeed attached to a specific note, the direction element can be associated with the note element that follows it in score order that is not in a different voice."_. You just attaches the pending dynamics mark to the next analyzed note without checking for the voice, but as dynamics marks are not voice related, I think it is OK to attach the AuxObj to the next note and it is not necessary to check for voice. In any case, although the `<direction>` element has an optional voice element, I've not seen it on the MusicXML samples. Also the voice elements is currently not parsed in `DirectionMxlAnalyser` (line 2638). Anyway, not necessary.

The only issue I see in your proposed code is related to the `clear_pending_note_rest_aux_objs()` method. If the dynamics mark cannot be attached to a note, instead of deleting it, it should be attached to the ImoDirection. But. how to do it? One possibility is to initially attach the AuxObj to the parent ImoDirection, and later, if it is decided to attach it to a note, find its current parent, remove the AuxObj from it and attach it to the note. Also it would be nice to delete the ImoDirection if it doesn't have more attached objects. And if the AUxObj is finally not attached to a note, it will be already properly attached to an ImoDirection in the right place.

In conclusion, it would be necessary to:
- always attach the AuxObj to the parent (DynamicsMxlAnalyser , line 2805)
```
    pSO->add_attachment(pDoc, pImo);
    if (!pSO->is_note_rest())
        m_pAnalyser->add_pending_note_rest_aux_obj(pImo);
```
- In `MxlAnalyser::add_to_pending_aux_objs()`: before attaching the AuxObj to the note, find its current parent, remove the AuxObj from it and attach it to the note. And remove from the model and delete the ImoDirection if it doesn't have more attached objects.

- In `clear_pending_note_rest_aux_objs()` just clear the vector, but do not delete the AuxObjs.

- Finally, I would suggest change some names:
    - `vector<ImoAuxObj*>  m_pendingNoteRestAuxObjs;`  -->  `vector<ImoDynamicsMark*>  m_pendingDynamicsMarks;`
    - `add_pending_note_rest_aux_obj()` --> `add_pending_dynamics_mark()`
    - `add_to_pending_aux_objs()` --> `attach_pending_dynamics_marks()`
    - `clear_pending_note_rest_aux_objs()` --> `clear_pending_dynamics_marks()`, or better remove the method and just do `m_pendingDynamicsMarks.clear();`

Hope I have not missed anything!

By the way,  I see that you are working to improve the scores layout, so to not overlap work just to inform you that  I'm currently studying better approaches to layout slurs to improve its appearance.

dmitrio95:
Thanks for your comments! I updated the pull request with the proposed changes, except that I haven't yet implemented removing empty `ImoDirection` objects. If I understand it correctly, `<direction>` tag (represented by `ImoDirection` object) may have several children, and it may theoretically happen that it has both dynamics mark and some spanner endpoint as its children. In this case it would be dangerous to delete the `ImoDirection` object as the spanner would only get added to it after its second endpoint is encountered. I wasn't able to find examples with dynamics, but there are cases when this happens for `<words>` and (not currently imported by Lomse) `<dashes>` elements, for example in [this test score](https://github.com/lenmus/lomse/blob/2918eb53ac60b4162b703e37340054ef817fc11b/test-scores/50047-cross-staff-beamed-group-more-space.xml#L2718). A possible solution may be to store a list of such potentially empty `ImoDirection` objects in `MxlAnalyzer` and check presence of attached objects for them after parsing the entire MusicXML file. If this is fine, I can add this logic to this pull request as well.

cecilios:
Yes, your understanding of `ImoDirection` is correct. It can have several AuxObjs and RelObj attached.

My explanation was not clear enough because when I was saying _"it doesn't have more attached objects"_ I was referring to both AuxObjs and RelObjs. So, as you correctly guessed the `ImoDirection` parent must be deleted only if the dynamics mark auxobj is the only attachment. 

The internal model is not very well designed and has some nuisances. I should work on cleaning it a little but I always have other things to do! All attached ImoAuxObjs are not direct children. Instead the parent has a child ImoAttachments and the AuxObjs are attached there as children. As to RelObjs, it is similar, and they are attached to a child ImoRelations object, but in this case the attached RelObjs instead of being children of ImoRelations, they are included in an internal std::list and are not part of the tree. 

You should deal with this as follows:

```
void MxlAnalyser::attach_pending_dynamics_marks(ImoNoteRest* pNR)
{
    for (ImoDynamicsMark* pDynamics : m_pendingDynamicsMarks)
    {
        ImoObj* pOldParent = pDynamics->get_parent_imo();       //this is the ImoAttachments child in the parent ImoContentObj (the ImoDirection obj)
        pOldParent = pOldParent->get_parent_imo();     //this is the ImoContentObj (ImoDirection)   

        if (pOldParent)     //Must exist! Otherwise big big bug or memory corruption! Probably no need to check this
        {
            pOldParent->remove_attachement(pDynamics);

            ImoStaffObj* pSO = dynamic_cast<ImoStaffObj*>(pOldParent);
            if (pSO && pSO->get_num_attachments()==0 && !pSO->has_relations())   //is empty?
            {
                ImoObj* pMusicData = pSO->get_parent_imo();     //this is the ImoMusicData parent
                pMusicData->remove_child_imo(pSO);
                delete pSO;
            }
        }

        pNR->add_attachment(m_pDoc, pDynamics);
    }
}
```
I have not tested this code but I think it is correct. If you would like to try, it would be nice if you could prepare two minimal MusicXML scores, one with the `<dynamics>` followed by a `<note>` and the other one with the `<dynamics>` followed by the final barline, to add them to the unit tests (in `lomse_test_mxl_analyser.cpp` after line 1714) and check both behaviors. But if you don't feel sure or other, don't worry I'll do it later as part of the PR test.

Thank for contributing!

dmitrio95:
> My explanation was not clear enough because when I was saying _"it doesn't have more attached objects"_ I was referring to both AuxObjs and RelObjs.

Yes, I see, but my point is that on reading spanners only `*Dto` objects are created, and real `ImoRelObj`s may be created and attached to `ImoDirection`s (or any other objects) later than the next note is encountered in the document. Therefore it is not possible to reliably check at that point whether `ImoDirection` really has no other attachments or it is just waiting for some `RelObj` to be attached. If so, this check can possibly be postponed to the end of reading MusicXML document, in this case it should not cause such issues.

cecilios:
Sorry! You are right. You have to postpone the checking and deletion or, better, ignore this. And empty `ImoDirection` doesn't cause any problem; it was just optimization. In any case, if desired, in future  they can be filtered out and deleted when the score is finished, when building the ColStaffObjs.

src/parser/mxl/lomse_mxl_analyser.cpp
```diff
@@ -7765,6 +7776,23 @@ void MxlAnalyser::clear_pending_relations()
...
     m_lyricIndex.clear();
+    m_pendingDynamicsMarks.clear();
+}
+
+//---------------------------------------------------------------------------------------
+void MxlAnalyser::attach_pending_dynamics_marks(ImoNoteRest* pNR)
+{
+    for (ImoDynamicsMark* pDynamics : m_pendingDynamicsMarks)
+    {
+        ImoObj* pOldParent = pDynamics->get_parent_imo();
```
cecilios, line 7787: This line needs to be changed to
```
ImoObj* pOldParent = pDynamics->get_parent_imo();       //this is the ImoAttachments child in the parent ImoContentObj (the ImoDirection obj)
pOldParent = pOldParent->get_parent_imo();     //this is the ImoContentObj (ImoDirection)   
```

cecilios:
Don't worry about the unit tests. I'll do it after testing and merging the PR.

src/parser/mxl/lomse_mxl_analyser.cpp
dmitrio95: As far as I understand, dynamics marks are currently children of `ImoAttachments` in the internal model tree so this code should currently do a correct thing detaching dynamics mark from its parent `ImoAttachments`.

I agree however that it would be better to remove dynamics from `ImoDirection` by something like `remove_attachment()` call, thus making it symmetrical to the way it has been added there. However I see two issues here:
1. `remove_attachment()` function [deletes](https://github.com/lenmus/lomse/blob/d4537fb5e2c5397ffdfd50504364897d00be0094/src/internal_model/lomse_internal_model.cpp#L2153) the object it detaches. This is not what we want here so something different would be needed for this purpose (I see few `remove_but_not_delete...` functions across the code, possibly something like this would work here).
2. Finding `ImoDirection` by two `get_parent_imo()` is less obvious to understand when reading the code, and relevant changes in the internal model structure may break this code in a way which will be easy to miss. Therefore I think it would be preferable to use something like [`get_contentobj_parent()`](https://github.com/lenmus/lomse/blob/d4537fb5e2c5397ffdfd50504364897d00be0094/src/internal_model/lomse_internal_model.cpp#L662) in such context.

For now I have pushed [these changes](https://github.com/lenmus/lomse/compare/86150bb4427063e006aeb0ac1bbe799dcfddf5a9..1009add1961253d554f17a617081612e9ea6b98e) which should address the listed considerations.

cecilios: Thank you very much for your effort to improve the code. Both changes, `get_contentobj_parent()` and `remove_but_not_delete_attachment()` are certainly very well thought. The internal model needs some refactoring. Perhaps the next task I should do is to start with it.

cecilios:
I think that now your PR is ready to be merged. I will do some tests and, if nothing detected, I will merge it.

Thank you!

By the way, I am really impressed by your ability to navigate the code and quickly learn its structure without needing any help. Congratulations! Nevertheless, if at any moment you would like to comment anything or would like clarifications on something, please feel free to ask or to send me a PM. I'll be glad to help you!

cecilios:
I've tested this PR and although it works as expected and really improves the positioning of dynamics marks, one of the regression tests (test score 50046) produced a result that I didn't expected. This is was was rendered before this PR:

50046-cross-staff-beamed-group-more-space-1

And this after merging the PR:

50046-new
And the reference (original pdf score) is:
DebuMandSample-reference-pdf

As expected, dynamics marks 'sfp' and 'pp' are now centered on the notes. But the unexpected result is the position change of 'pp' dynamics mark. To understand the reason, first let's analyze the MusicXML code. Ignoring fermatas and other not relevant elements the definition order is:
```
    Part 2, Measure 1:
    - The grace note is defined in voice 1 but in staff 2 (!)
    - A direction to start a wedge, below staff 1
    - A direction in staff 1, positioned below staff 1, with the 'sfp' dynamics mark --> [1] See below
    - Note G4 half dotted, in staff 1
    - A direction to finish the wedge at end of measure
    - A backup to start of measure
    - the rest

    Part 2, Measure 2:
    - attributes: new clef G2
    - note chord G3 D4 A4 eighth voice=2 staff=2
    - direction in staff 2, positioned above staff 2, dynamics 'pp'  ---> [2] See below
    - note chord G4 D5 A5 in staff 1
```

[1] 'sfp' dynamics mark:
- before PR: direction in staff 1, positioned below staff 1 ==> placed below staff 1. OK
- after PR: transferred to note G4 in staff 1 and placed bellow it ==>  placed below staff 1. OK

[2] 'pp' dynamics mark:
- before PR: direction in staff 2, positioned above staff 2 ==> INCORRECTLY placed below staff 2. Should be above staff 2
- after PR: transferred to G4 note in staff 1 and thus, placed above it ==> INCORRECTLY placed below staff 1. Should be above staff 1 (notice that the transfer changes the expected position! ) 

My preliminary conclusions:

- There is a bug in MusicXML importer in line 2818 that causes that all directions will be placed in the default position. I didn't have time today to investigate if the direction position is transferred to its content. But it seems that all directions are rendered bellow the staff. Don't worry about this, I will fix this bug and apply it after merging your PR.

- Your PR does not take into account that the direction is defined in one staff and the note that receives the attachment can be in a different staff. Therefore, during the transfer, the position of the dynamics mark should be adjusted.

- The  'sfp' dynamics mark is defined at the same time position than the wedge; therefore, they are vertically stacked. The only way of getting the original layout (the 'sfp' before the wedge) is by accepting user defined positions (default-x' and default-y' attributes). But this is something that I'm reluctant to import.

I have to think a little more about all this, but I would like to share with you my current thoughts. Let's think a little about this findings and then decide what to do.

dmitrio95:
>  Your PR does not take into account that the direction is defined in one staff and the note that receives the attachment can be in a different staff. Therefore, during the transfer, the position of the dynamics mark should be adjusted.

Indeed, that should probably be fixed. It should be easy to correct this by checking staff number like it is already done for octave shifts. However doing so will attach the "pp" dynamics mark in this example to the second chord in the second staff which seems to be even more wrong in the context of that score. Perhaps we should just leave such dynamics attached to their original `ImoDirection`s if they cannot be attached directly to the next note encountered in the score? Not sure whether this makes sense in a general case, but at least for this example it will work better (although this MusicXML does not seem to fully correspond to the initial score anyway).

> The 'sfp' dynamics mark is defined at the same time position than the wedge; therefore, they are vertically stacked. The only way of getting the original layout (the 'sfp' before the wedge) is by accepting user defined positions (default-x' and default-y' attributes). But this is something that I'm reluctant to import.

From my point of view a good option here would be to automatically align a wedge if it starts or ends with a dynamics mark, so in this example the wedge would be shortened from its left side to make a room for "sfp" mark to appear below the note. This is exactly what the reference picture shows, and this is something that MuseScore does by default in such situations (I was even implementing some parts of this logic there :) ).

Correct `default-x` values will likely not be provided in MusicXML documents in most cases (and in this example the relevant elements don't have them too) so importing user-defined position will probably not resolve this issue for such scores.

cecilios:
> Perhaps we should just leave such dynamics attached to their original ImoDirections ...

No, I think it is better to attach it to the notes/rests unless some ad hoc layout rule, such as the one suggested for the spf and wedge case, applies. The normal procedure could be to attach to next note without considering the staff, but to change dynamics placement to try to keep the position, when note staff != direction staff:
```
                                                     final
case            direction   note    placement       placement
  1   staff       1          2       above            ?        -> above 1
  2   staff       1          2       below           above     -> above 2 = below 2
  3   staff       2          1       above           below     -> below 1 = above 2
  4   staff       2          1       below            ?        -> below 2

```
As you can see there are some issues to solve: 
- new placement codes, for cases 1 and 4 
- instruments with more than 2 staves (organ?)

So, for now perhaps the best solution is to merge the PR and, in future, try to deal with this. There are more urgent things to do now.

As to the 'sfp' dynamics mark defined at the same time position than the wedge:

> a good option here would be to automatically align a wedge if it starts or ends with a dynamics mark

I also thought about this solution and I thing it is the best approach. But in any case, this has nothing to do with this PR, it is not urgent, and can be solved later.

> MuseScore does by default in such situations

I didn't know. I never look at the solutions in other projects code. I like to think and to find solutions by myself, to learn. That takes more time but I do not have deadlines for having the code ready  :-)

>  I was even implementing some parts of this logic there :)

That's great! when you send the first PR I noticed that you also contributes to MuseScore and I was surprised that you would like to contribute to Lomse. But tank you and welcome! MuseScore is a very nice project and I use it for creating scores. In 2012 I decided not to pay for web hosting for my project and Thomas Bonte contacted me and offered to host my project. We were taking about collaboration in MuseScore but our projects were so different that to do it I had to abandon mine. So, I decided to continue with my project. For you, having studied the internals of both projects will have given you a much better knowledge of this area of dealing with music notation! By the way, Lomse offers much more than it is required for my application, and thus I develop Lomse just for fun and to learn. I choose issues to solve without any criteria, just my personal taste at the moment. Therefore, if you would like that I give priority to some issues, please tell me. 

Well, returning to the PR I think that it should be merged as it is. Agree? Then please rebase and I'll do it.

dmitrio95:
Great! I had a chance to meet Thomas in person, as well as the other MuseScore founders, Werner and Nicolas. I didn't know that you also know each other, the world is that small :)

As for Lomse, I use it for my personal project I have started recently, which is basically a score reader which tries to listen to what a human plays and track a current position in a score to always show the relevant pages. The music notation library from MuseScore would be less convenient to use in such project since it is heavily dependent on Qt and has a more restrictive license (the latter is not that relevant right now since I didn't publish my project anyway, but that may matter if I ever decide to). Therefore I am mostly dealing with issues which I see in the context of my project, but I will certainly be interested in making Lomse better in general. I don't also think that it currently makes sense to prioritize any issues relevant solely to my project: I think it would be much more useful for a project like Lomse to improve its different areas rather than focus on one particular way of using the library. Maybe it could be useful though if I report some of those issues I encounter: currently I somehow tend to report them only after figuring out a way to fix them.

> Well, returning to the PR I think that it should be merged as it is. Agree? Then please rebase and I'll do it.

OK, I'll push a rebased version of the pull request now.

cecilios:
Thank you!

Yours is a very interesting project, and great to avoid turning pages! But audio analysis must be the toughest part, specially if not a single instrument!

PR #295, opened 2021-07-09
Draw repeat barline dots on all staves

dmitrio95:
Currently Lomse draws dots of repeat barlines only on the first staff of the instrument, for example:
before

However repeat dots are commonly placed on each staff in multi-staff scores. This PR makes repeat dots appear on all staves, so the example shown above now looks like the following:
after

As repeat dots drawing is handled by a barline shape, this shape needs to know coordinates of all staves in its range, so most of the changes in this PR are about forwarding this information to barline shapes. The first commit here reverts b3c2f2b4c2d32040fad74b8f3efaa91a2175ad62 and instead ensures that vertical staff position shifts are reset when a new system's layout begins. This is needed to ensure that `InstrumentEngraver::get_top_line_of_staff()` (which definition is changed to take these shifts into account, consistently with the other relevant methods of `InstrumentEngraver`) always gives correct positions for staff coordinates in the current system.

### DIFF
--- a/include/lomse_barline_engraver.h
+++ b/include/lomse_barline_engraver.h
@@ -39,6 +39,7 @@ namespace lomse
 
 //forward declarations
 class ImoBarline;
+class InstrumentEngraver;
 class GmoBoxSliceInstr;
 class GmoShape;
 class ScoreMeter;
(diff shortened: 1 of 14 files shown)


cecilios:
Thank you for this PR. I will check it as soon as possible.

cecilios:
I have reviewed the PR and all the changes make sense to me. All unit tests and regression test passed, so I'll merge it.

I had doubts about whether the static method in `BarlineEngraver` should be defined in other place. It contains general knowledge about barlines and should perhaps be placed in a new `BarlineUtilities` class defined in `lomse_score_utilities.h`, but my second thought was that this method will probably never be used, other than this specific case in `BarlineEngraver`. Thus, it is simpler to keep it as is and, if convenient, it could be moved in future.

Thanks again for this PR!

PR #296, opened 2021-07-11
MusicXML: fix wedge endpoints order if wedge end is found before wedge start

dmitrio95:
Similar to slurs (which are handled in #286), wedge end can be found before wedge start in a MusicXML file. The code responsible for reading wedges has a provision for this case so such wedges are already imported. However order of endpoints of such wedges is wrong which leads to a crash on an attempt to render such wedge. A minimal example of this case can be given by this file: 
[wedge.musicxml.zip](https://github.com/lenmus/lomse/files/6796796/wedge.musicxml.zip)

This pull request changes handling of wedges' endpoints in the same way as it has been done for slurs in #286, so order of wedge endpoints is corrected so that wedge stop always goes last. This allows to properly import and display such wedges.

This pull request though doesn't deal with the reason of the crash itself, so an incorrect MusicXML file still can lead to a crash if trying to engrave such score. So in the future some fix for this crash could still be useful (for example, by not assuming a correct order or endpoints on engraving stage or by validating endpoints order after reading a MusicXML file).

### DIFF
--- a/src/parser/mxl/lomse_mxl_analyser.cpp
+++ b/src/parser/mxl/lomse_mxl_analyser.cpp
@@ -7399,7 +7399,7 @@ protected:
 
         if (m_pAnalyser->wedge_id_exists(num) && value != "continue")
         {
-            m_pInfo1->set_start(false);
+            m_pInfo1->set_start(value != "stop");
             int wedgeId =  m_pAnalyser->get_wedge_id_and_close(num);
             m_pInfo1->set_wedge_number(wedgeId);
             if (value == "crescendo")
@@ -7407,7 +7407,7 @@ protected:
         }
         else if (value == "crescendo" || value == "diminuendo" || value == "stop")
         {
-            m_pInfo1->set_start(true);
+            m_pInfo1->set_start(value != "stop");
             int wedgeId =  m_pAnalyser->new_wedge_id(num);
             m_pInfo1->set_wedge_number(wedgeId);
             m_pInfo1->set_crescendo(value == "crescendo");
@@ -7822,7 +7822,7 @@ void MxlAnalyser::add_relation_info(ImoObj* pDto)
     else if (pDto->is_volta_bracket_dto())
         m_pVoltasBuilder->add_item_info(static_cast<ImoVoltaBracketDto*>(pDto));
     else if (pDto->is_wedge_dto())
-        m_pWedgesBuilder->add_item_info(static_cast<ImoWedgeDto*>(pDto));
+        m_pWedgesBuilder->add_item_info_reversed_valid(static_cast<ImoWedgeDto*>(pDto));
     else if (pDto->is_octave_shift_dto())
         m_pOctaveShiftBuilder->add_item_info(static_cast<ImoOctaveShiftDto*>(pDto));
 }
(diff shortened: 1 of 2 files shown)


cecilios:
Thanks for this PR.

I will review it as soon as possible.

cecilios:
Thank you for this PR.

I've reviewed the changes and they are OK for me. Unit test passes as well as regression tests, so I will merge this.

As to your point:
> This pull request though doesn't deal with the reason of the crash itself, so an incorrect MusicXML file still can lead to a crash if trying to engrave such score.

I'm not sure what you mean by “fixing the reason of the crash itself”. The cause of the crash is incorrect order of start/end points in lomse internal model. In lomse, spanners are modeled as `ImoRelObjs` attached to `ImoStaffObjs` (e.g. notes), and the start and end of the spanner are always in music time order (time position). Thus, the note (`ImoStaffObj`) to which the end of a spanner is attached must have a time position equal or greater than the note to which the start of the spanner is attached. Thus, as this is now fixed in MusicXML importer the last cause of the crash is already fixed.

Nevertheless, the crash takes place during engraving. When an `ImoRelObj` is engraved, if it is the end of the spanner lomse invokes `ShapesCreator::finish_engraving_relobj()`. Lomse assumes (by the internal model constrains explained above) that the start of the wedge has already been processed and, thus, that a `WedgeEngraver` object has been created. So, in `ShapesCreator::finish_engraving_relobj()` it is assumed that `pEngrv` is not `nullptr`.

Checking there (and probably in other places) for nullptr would fix the crash. But I'm not sure if this is needed. The important point is to ensure the validity of the internal model: than the start object of a spanner is defined before its end object. And this should be the responsibility of all importers and, thus, of MusicXML importer. As this PR fixes this a malformed MusicXML file will not cause a crash.

Nevertheless, if really required for robustness, I see two options:

1. The best one: a check could be added in `lomse_model_builder.cpp` (an additional step in `ModelBuilder::structurize()` ). This check could also fix any issue by reversing the start/end positions when incorrectly defined. But forces to do more computational work for any operation requiring to rebuild the internal model, and IMO this burden should be only in importers.

2. A quick and probably more simple fix/bypass: Checking in `ShapesCreator::finish_engraving_relobj()` (and probably in other places) that `pEngrv` is not nullptr.

But I would not do anything for now unless new crashes with MusicXML files are detected. And in those cases I would first explore the feasibility of fixing the importer and making it more robust against malformed files.

Is there something that I missed?

dmitrio95:
> As this PR fixes this a malformed MusicXML file will not cause a crash.

Actually this PR fixes the case when a valid MusicXML file causes a crash. The importer still relies on `<wedge>` tag attributes to sort wedge endpoinds, and if a MusicXML file marks wedge endpoints in an incorrect order then the crash will occur. For example, consider this file ([wedge_wrong.musicxml.zip](https://github.com/lenmus/lomse/files/6804720/wedge_wrong.musicxml.zip)): this is just the previous example but with `diminuendo` and `stop` being swapped for the wedge number 2. This MusicXML file is incorrect since those attributes should also follow music time order, but it would anyway be better not to crash because of it. So maybe some sort of validation after or during the import would indeed help.

cecilios:
>  if a MusicXML file marks wedge endpoints in an incorrect order then the crash will occur.

Thank you. You are right and the sample is illustrative. I will study in more detail the options to avoid crashes and implement something to fix this.

PR #297, opened 2021-07-11
Fix -Wdeprecated-copy compiler warnings from GCC

dmitrio95:
This PR fixes [`-Wdeprecated-copy`](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html) warnings which appear when compiling Lomse with GCC 10. The reason of the warnings is that some classes contain declarations only for one of copy constructor or copy assignment operator, but the other declaration was missing (but actually used by the library's code). Therefore a compiler is forced to generate the missing constructor/operator itself but this is deprecated in such situation since C++11 and actually should not normally happen since this violates [the rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)). Therefore in order to avoid the warning (and potential issues related to this situation) copy constructor and copy assignment operator should either both be user-defined (or deleted) or not defined at all so both their definitions will be defined by a compiler if needed.

In Lomse code the warnings result from manually defining copy constructor or assignment in two ways:
1) Exactly that way as it would be defined by a compiler (that is, just copying/assigning all members of this class and invoking base classes' copying/assignment if needed)
2) Defining them *almost* that way: copy constructors for `ImoLineStyle`, `ImoSystemInfo` and `ImoPageInfo` differ from the default definition in parameters of the base class constructor, but this doesn't seem to be necessary from these classes' semantics, and these copy constructors are actually never used (which can be verified by deleting these copy constructors with `= delete`).

Since the copy constructors or assignment operators in question are either identical to the default ones or unused, I went for removing them in both cases, so they can be defined by a compiler in the default way. This fixes the related warnings and should hopefully make semantics of these operations more clear.

The only case which is not covered by this PR is about [`PathAttributes`](https://github.com/lenmus/lomse/blob/0ff4802acba463bbbd259727c4eefe61c773f7e7/include/lomse_path_attributes.h#L167) class which needs to have user-defined copy constructor and assignment operator (since it has to copy its `fill_gradient` by creating a new instance of that gradient) but is used only in `agg::pod_bvector` container which does not honor destructor of this class. However I wonder whether `pod_bvector` can be replaced there by `vector` or `deque` from STL, so this class could manage lifetime of that gradient without relying on other objects to delete it?

### DIFF
--- a/include/lomse_staffobjs_table.h
+++ b/include/lomse_staffobjs_table.h
@@ -181,15 +181,6 @@ public:
                 }
             }
 
-            virtual ~iterator() {}
-
-            iterator& operator =(const iterator& it) {
-                m_pCurrent = it.m_pCurrent;
-                m_pNext = it.m_pNext;
-                m_pPrev = it.m_pPrev;
-                return *this;
-            }
-
 	        ColStaffObjsEntry* operator *() const { return m_pCurrent; }
 
             iterator& operator ++() {
--- a/include/lomse_score_iterator.h
+++ b/include/lomse_score_iterator.h
@@ -54,8 +54,6 @@ protected:
 
 public:
     StaffObjsIterator(ColStaffObjs* pColStaffObjs);
-    StaffObjsIterator(const StaffObjsIterator& it);
-    ~StaffObjsIterator();
 
 
     inline bool is_first() { return m_it == m_pColStaffObjs->begin(); }
--- a/include/lomse_xml_parser.h
+++ b/include/lomse_xml_parser.h
@@ -60,7 +60,6 @@ protected:
 
 public:
     XmlNode() {}
-    XmlNode(const XmlNode& node) : m_node(node.m_node) {}
     XmlNode(const XmlNode* node) : m_node(node->m_node) {}
 
     string name() { return string(m_node.name()); }
--- a/include/private/lomse_internal_model_p.h
+++ b/include/private/lomse_internal_model_p.h
@@ -3228,20 +3228,6 @@ protected:
     {
     }
 
-    ImoLineStyle(ImoLineStyle& info)
-        : ImoSimpleObj(k_imo_line_style)
-        , m_lineStyle( info.get_line_style() )
-        , m_startEdge( info.get_start_edge() )
-        , m_endEdge( info.get_end_edge() )
-        , m_startStyle( info.get_start_cap() )
-        , m_endStyle( info.get_end_cap() )
-        , m_color( info.get_color() )
-        , m_width( info.get_width() )
-        , m_startPoint( info.get_start_point() )
-        , m_endPoint( info.get_end_point() )
-    {
-    }
-
 public:
     virtual ~ImoLineStyle() {}
 
@@ -3621,7 +3607,6 @@ protected:
     friend class ImoDocument;
     friend class ImoScore;
     ImoPageInfo();
-    ImoPageInfo(ImoPageInfo& dto);
 
 public:
     virtual ~ImoPageInfo() {}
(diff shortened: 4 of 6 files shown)


include/lomse_staffobjs_table.h
```diff
@@ -181,15 +181,6 @@ class ColStaffObjs
                 }
             }
 
-            virtual ~iterator() {}
```
dmitrio95, line 184: This change is not exactly needed to fix this warning but there doesn't seem to be a reason to inherit from this iterator class, and no class in Lomse does it. Hence there is no need for a virtual destructor here so I have removed it as well along with the explicit `operator=` definition.

cecilios:
Thanks for this PR.

I will review it as soon as possible and answer your question about replacing `pod_bvector` by STL containers.

include/lomse_staffobjs_table.h
cecilios: I started long ago with Lomse and by that time I had experience with C but little experience and knowledge of C++. I new about the `rule of three` and the improved version, the `rule of five`, much later but still I'm not used to take this rule into account when defining classes and also I never reviewed the code to comply with it. Probably you will see many definitions that need fixes. I usually define destructors as `virtual` without more considerations, just in case I derive for that class.

Also, another problem I have is in proper usage of `const` qualifier. Whenever I try to use it I find problems that force me to duplicate definitions with different `const` combinations, and, usually I opt by the shortest path: to remove the `const` qualifier and leave just one definition valid for all cases.

Please, fell free to fix these issues when you find any, without an specific PR for the fix.

cecilios:
Thanks for this PR and your explanations.

> The only case which is not covered by this PR is about PathAttributes class which needs to have user-defined copy constructor and assignment operator (since it has to copy its fill_gradient by creating a new instance of that gradient) but is used only in agg::pod_bvector container which does not honor destructor of this class. However I wonder whether pod_bvector can be replaced there by vector or deque from STL, so this class could manage lifetime of that gradient without relying on other objects to delete it?

You are right that deleting `GradientAttributes` in `BitmapDrawer::delete_paths()` is quite dirty. Also there is a risk of memory leaks if `PathAttribute` is not used to render something. But your suggestion of replacing `pod_bvector` by an STL container will require to analyse how `AttrStorage` is used and to ensure that `agg` methods expecting a `pod_bvector` will receive the right container. I do not fell confident to do that analysis now but feel free to do it and to fix this dirty code by using an STL container or with any other solution.

Meanwhile, as a bypass to fix the compiler warning, I would suggest to just remove the destructor and move the comment to the line allocating the `GradientAttributes` object, something as:

```
{
    if (attr.fill_gradient)
        fill_gradient = LOMSE_NEW GradientAttributes(attr.fill_gradient);

    //AWARE: fill_gradient will be deleted **only** if this PathAttributes
    //       object is used to render something. It will be deleted in
    //       BitmapDrawer::delete_paths().
    //TODO: Fix this to avoid memory leaks if PathAttributes is not used
}
```

The changes looks OK to me, unit tests pass and regression tests also pass so it is OK to me to merge this PR. But I will wait in case you have any comments.

dmitrio95:
Thanks for your comments! I will review this case once more and try to replace the container if possible. From what I have briefly seen this `pod_vector` is not used in any AGG-specific API but I could possibly miss something here.

> Meanwhile, as a bypass to fix the compiler warning, I would suggest to just remove the destructor and move the comment to the line allocating the GradientAttributes object, something as:

This warning does not consider destructors (because of a large number of "false positive" warnings if destructor is empty or something like that), and the warning in this case is about `PathAttributes` having user-defined copy constructor but not a copy assignment operator. I couldn't neither remove the constructor nor define the `operator=` in a way which would be logical (with copying its `GradientAttirbutes`) since both these options break the current `PathAttributes` semantics. That's why I didn't do any changes about it.

Anyway it would probably be better to handle `PathAttributes` separately so feel free to merge this PR if the current changes are OK.

cecilios:
Ok, thank you. I'm going to merge this.

Issue #218, opened 2019-09-27
Support pedal marks


cecilios:
From #117. Currently, pedal marks are not supported in Lomse and, therefore, not imported from MusicXML files.

This should be fixed.

PR #299, opened 2021-08-01
Add initial support for arpeggio

dmitrio95:
This pull request contains a basic implementation of arpeggio in Lomse. This implementation is somewhat limited (which I have briefly described below) but it is still able to import an arpeggio attached to one chord and display it when rendering a score. Therefore it can be useful to handle arpeggio in the most common cases and is hopefully able to serve as a basis for further improvements.

## Examples

Here are some examples of arpeggio rendering with the current version of this PR:
 - [`32d-Arpeggio.xml`](https://github.com/cuthbertLab/musicxmlTestSuite/blob/main/xmlFiles/32d-Arpeggio.xml) from the [unofficial MusicXML test suite](https://github.com/cuthbertLab/musicxmlTestSuite) (non-arpeggiate is not implemented in this pull request):
arpeggio_testsuite
 - Arpeggio of different lengths:
arpeggio_test_length
 - Arpeggio on cross-staff chords (based on [this example](https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/tutorial-apres-un-reve), extra dynamic marks were added to test the case when staff distance gets increased in the course of engraving):
arpeggio_cross_staff_chord
 - Some real-world examples:

| Example 1  | Example 2 | Example 3 |
| ------------- | ------------- | ----------- |
| arp1 | arp2 | arp3

## Limitations of the current implementation

In this PR arpeggios are implemented in the internal model similarly to chords, as `RelObj`s connecting several notes. This seems to be natural in the context of the Lomse score model and potentially allows representation of a "true" cross-staff arpeggio involving separate chords on different staves. However reading such arpeggios is not implemented in the MusicXML importer (primarily because it wasn't clear for me how they are supposed to be represented in a MusicXML file), and the importer assumes that arpeggio always spans only for one chord. Similarly, the engraver only checks for arpeggio in one note and renders it only for one chord. This works fine in the most common case of a one-chord arpeggio but in order to support cross-staff arpeggio this will need to be corrected.

Also this PR does not include support of arpeggio in the LDP format (not sure if I need to do something special to add it or it is handled automatically?) and does not include support for MIDI rendering of arpeggio so all chords would sound non-arpeggiated on playback.

### DIFF
--- a/src/graphic_model/lomse_box_slice_instr.cpp
+++ b/src/graphic_model/lomse_box_slice_instr.cpp
@@ -169,7 +169,7 @@ void GmoBoxSliceStaff::reposition_shapes(const vector<LUnits>& yShifts,
             else if ((*it)->is_shape_note())
             {
                 GmoShapeNote* pShapeNote = static_cast<GmoShapeNote*>(*it);
-                LUnits increment = (yShift + yShifts[m_idxStaff-1]);
+                LUnits increment = (yShift - yShifts[m_idxStaff-1]);
                 pShapeNote->reposition_shape(yShift);
 
                 if (pShapeNote->is_cross_staff_chord())
@@ -177,6 +177,11 @@ void GmoBoxSliceStaff::reposition_shapes(const vector<LUnits>& yShifts,
                     if (pShapeNote->is_chord_start_note())
                     {
                         pShapeNote->increment_stem_length(increment);
+
+                        GmoShapeArpeggio* pArpeggio = pShapeNote->get_base_note_shape()->get_arpeggio();
+
+                        if (pArpeggio)
+                            pArpeggio->increase_length_up(increment);
                     }
                     else if (pShapeNote->is_chord_flag_note() && !pShapeNote->is_up())
                     {
--- a/include/lomse_bitmap_drawer.h
+++ b/include/lomse_bitmap_drawer.h
@@ -197,6 +197,7 @@ public:
     int draw_text(double x, double y, const std::string& str) override;
     int draw_text(double x, double y, const wstring& str) override;
     void draw_glyph(double x, double y, unsigned int ch) override;
+    void draw_glyph_rotated(double x, double y, unsigned int ch, double rotation) override;
 
 
     //copy/blend a bitmap
--- a/include/lomse_calligrapher.h
+++ b/include/lomse_calligrapher.h
@@ -70,10 +70,12 @@ public:
     int draw_text(double x, double y, const wstring& str, Color color,
                   double scale=1.0);
     void draw_glyph(double x, double y, unsigned int ch, Color color, double scale);
+    void draw_glyph_rotated(double x, double y, unsigned int ch, Color color, double scale, double rotation);
 
 protected:
     void draw_glyph(double x, double y, unsigned int ch, Color color);
     void set_scale(double scale);
+    void set_scale_and_rotation(double scale, double rotation);
 
 };
 
@@ -93,6 +95,7 @@ public:
 
     LUnits measure_width(const std::string& str);
     LUnits measure_width(const wstring& str);
+    LUnits get_advance_x(unsigned int ch);
     LUnits get_ascender();
     LUnits get_descender();
     LUnits get_font_height();
--- a/include/lomse_chord_engraver.h
+++ b/include/lomse_chord_engraver.h
@@ -98,6 +98,8 @@ protected:
     ChordNoteData* m_pLinkNoteData;     //for cross-staff chords, the last note in the
                                         //same staff than the flag note
 
+    bool m_fSomeAccidentalsShifted;
+
 public:
     ChordEngraver(LibraryScope& libraryScope, ScoreMeter* pScoreMeter, int numNotes,
                   double fontSize, int symbolSize);
@@ -136,6 +138,7 @@ protected:
     void find_reference_notes();
     void layout_noteheads();
     void layout_accidentals();
+    void layout_arpeggio();
     void determine_stem_x_left();
     void add_stem_and_flag();
     void add_stem_flag_segment(StemFlagEngraver* engrv);
@@ -156,6 +159,8 @@ protected:
 
     LUnits check_if_accidentals_overlap(GmoShapeAccidentals* pPrevAcc,
                                         GmoShapeAccidentals* pCurAcc);
+
+    void ensure_note_bounds_updated();
 
     //helpers
     inline bool has_stem() { return m_fHasStem; }
--- a/include/lomse_drawer.h
+++ b/include/lomse_drawer.h
@@ -621,6 +621,7 @@ public:
         @param ch  The character to render.
     */
     virtual void draw_glyph(double x, double y, unsigned int ch) = 0;
+    virtual void draw_glyph_rotated(double x, double y, unsigned int ch, double rotation) = 0;
     //@}    //Text redering methods
 
 
--- a/include/lomse_engraving_options.h
+++ b/include/lomse_engraving_options.h
@@ -77,6 +77,8 @@ namespace lomse
 #define LOMSE_LEGER_LINE_OUTGOING        5.0f
 #define LOMSE_GRACE_NOTES_SCALE          0.60f  //Scaling factor for grace notes size
 #define LOMSE_CUE_NOTES_SCALE            0.75f  //Scaling factor for cue notes size
+#define LOMSE_ARPEGGIO_SPACE_TO_CHORD    6.0f   //Spacing between an arpeggio and a chord
+#define LOMSE_ARPEGGIO_MAX_OUTGOING      3.0f   //Amount of space arpeggio is allowed to go beyond a chord's top or bottom note
 
 //System layouter
     //spacing function parameters
--- a/include/lomse_glyphs.h
+++ b/include/lomse_glyphs.h
@@ -319,6 +319,10 @@ enum EGlyphIndex
 
     //Multi-segment lines (U+EAA0 - U+EB0F)
     k_glyph_trill_wiggle_segment,           // Trill wiggle segment
+    k_glyph_arpeggiato_wiggle_segment_up,   // Arpeggiato wiggle segment, upwards
+    k_glyph_arpeggiato_wiggle_segment_down, // Arpeggiato wiggle segment, downwards
+    k_glyph_arpeggiato_arrow_up,            // Arpeggiato arrowhead up
+    k_glyph_arpeggiato_arrow_down,          // Arpeggiato arrowhead down
 
     //Tremolos (U+E220 - U+E23F)
     k_glyph_tremolo_1,     // Combining tremolo 1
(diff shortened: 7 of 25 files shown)


cecilios:
Thank you for this PR. Great work! With this PR you understand the Lomse code almost as well as I do!

I'm currently on vacation, short on time for this and with limited access to internet. So reviewing the PR will take me some time. But I'll be happy to merge this as soon as I review and test the code, hopefully very soon.

After an initial review of the changes, my impression is that your approach for this PR is the one I would have followed. The limitations you mention in importing MusicXML arpeggios and on engraving are totally acceptable. You even took into account that the length of the arpeggio line could change if more space is added between staves later!

The lack of arpeggio support in playback is also something totally acceptable, to be addressed in future.

As for LDP, the short answer is: forget about LDP. The long answer is that the LDP format is not handled automatically but, in any case, LDP is only used in my application and I do not expect any more needs. In fact, the application is very old and I will stop maintaining it very soon. Therefore, LDP support could probably be totally removed in the near future. The LDP format is also used in Lomse as the format for the  stack of undo/redo operations for edition commands; therefore, to remove LDP from the library, this issue has to be addressed, as well as to replace the LDP format in many unit tests. As my application does not need score edition, and nobody has ever posted an issue related to this, I give no priority to edition commands and my efforts are on MusicXML, layout and playback, expanding the support for more notation. Forget about LDP.

There is something in your comments that leaves me intrigued: the issue of "true" cross-staff arpeggios. Are you referring to two independent chords, on different staves, joined by the same arpeggio line? Is this a valid notation? My books on music notation are at home and I cannot check this now; I will have to investigate this as well as the possible MusicXML support for this, if it exist. Do you have any sample scores displaying this? In any case, if the result of this research is that it is a valid notation that has to be supported, this could affect a lot to the solution implemented in this PR but, in any case, an improvement for future.

As said, my first impression is that you did an excellent work! Thank you.

src/graphic_model/lomse_box_slice_instr.cpp
```diff
@@ -169,14 +169,19 @@ void GmoBoxSliceStaff::reposition_shapes(const vector<LUnits>& yShifts,
             else if ((*it)->is_shape_note())
             {
                 GmoShapeNote* pShapeNote = static_cast<GmoShapeNote*>(*it);
-                LUnits increment = (yShift + yShifts[m_idxStaff-1]);
```
cecilios, line 172: Just by curiosity, how did you spot this bug? An issue in any sample score?

dmitrio95:
Thanks for your response! It would be totally fine for me to wait for any time so please feel free to make a review any time later.

> There is something in your comments that leaves me intrigued: the issue of "true" cross-staff arpeggios. Are you referring to two independent chords, on different staves, joined by the same arpeggio line? Is this a valid notation?

Yes, I don't remember seeing it often in scores I have dealt with, but this notation exists for the case when two hands should arpeggiate chords sequentially, as a one arpeggio. Some examples can be found, for example, in [Listzt's](https://imslp.org/wiki/Special:ReverseLookup/597) (page 9) or [Grieg's](https://imslp.org/wiki/Special:ReverseLookup/281000) works:
Screenshot_20210803_013515
Screenshot_20210803_013329
MusicXML is also intented to support this, as follows, for example, from [this issue](https://github.com/w3c/musicxml/issues/281) but at this moment I don't have any examples of MusicXML files containing cross-staff arpeggio.

> In any case, if the result of this research is that it is a valid notation that has to be supported, this could affect a lot to the solution implemented in this PR but, in any case, an improvement for future.

Actually I think it might be possible to implement this without very large changes to the structure introduced here. Possibly it would be enough to make a separate engraver for arpeggio which would get notified by chord engravers that they are ready for adding an arpeggio in the same way as chord engravers get triggered by the corresponding note engravers as these notes are found in a score. Arpeggio would then be engraved when its last chord is ready for that. Expanding cross-staff arpeggio if a staff distance changes is already implemented as a part of this PR so it would likely work as it is.

src/graphic_model/lomse_box_slice_instr.cpp
dmitrio95: This issue affects arpeggio at cross-staff chords, as I have reused this `increment` value for increasing arpeggio length at a staff distance change. Previously this value has only been used for expanding a stem segment between staves, and the extra length resulting from the `+` sign didn't result in any visible artifacts: the elongated stem segment just got drawn on top of another stem segment which doesn't produce any visible difference. With cross-staff arpeggio this results in an incorrect arpeggio length so I had to correct this expression to fix it.

cecilios:
No problems found neither in unit tests nor in regression tests. Also, no issues to comment in your code, so the PR is mergeable as it improves Lomse by adding partial support for arpeggios. I will do it.

I also have been thinking on how to model and deal with arpeggios, and also considering arpeggios that go cross staff across sets of chords (let's call them "cross-chords" arpeggios). You said that:

> Actually I think it might be possible to implement this without very large changes to the structure introduced here. Possibly it would be enough to make a separate engraver for arpeggio which would get notified by chord engravers that they are ready for adding an arpeggio in the same way as chord engravers get triggered by the corresponding note engravers as these notes are found in a score. Arpeggio would then be engraved when its last chord is ready for that. Expanding cross-staff arpeggio if a staff distance changes is already implemented as a part of this PR so it would likely work as it is.

The approach you followed to implement arpeggio lines is very interesting. And your ideas for expanding the solution to deal with cross-chords arpeggios looks feasible to me. I agree with the idea of creating a separate engraver for the arpeggio line to engrave it at current position, perhaps similar to the barline engraver for dealing with cross-staff spacing increments. This will decouple the engraver from the chord engraver and will facilitate dealing with cross-chords arpeggios. Triggering this engraver when the last chord in the arpeggio is engraved seems the only feasible approach. My only concern is about horizontal spacing. Currently, as the arpeggio line is engraved as part of the chord shape and the line shape is added to a chord note, the spacing algorithm automatically deals with its space, as it is part of the note shape. So the new engraver for the arpeggio line still needs to attach the arpeggio line to a note shape in the chord and recompute its boundings.

If I had implemented arpeggio lines I probably would have done it in a different way. As engraving chords is complex enough and arpeggio lines are just a type of line before the chord, I would not even considered the possibility of engraving the arpeggio lines as part of the chord engraving process. So probably I would have modeled "arpeggiated" as an attribute for chords (just for playback or other needs), as member variables in ImoChord object. And the arpeggio line modeled as an independent ImoStaffObj (ImoArpeggio, a non-timed staffObj) with information about the arpeggio line start and end notes; as they could be both in different chords it automatically will cope with cross-chords arpeggios. This approach will affect to the spacing algorithm (but just to add ImoArpeggio as another type of non-timed staffObj, no changes in computations needed) and will require an independent engraver, simple, just to engrave the line at current position, perhaps similar to the barline engraver for dealing with cross-staff spacing increments.

But your approach is also understandable, simple and valid, and unless complex problems detected when dealing with cross-chords arpeggios I would keep it.

Having reviewed the MusicXML approach for arpeggios and cross-chord arpeggios, I really find it as something requiring improvements, so I have posted my thoughts here: https://github.com/w3c/musicxml/issues/423.

dmitrio95:
Thanks for merging! Actually I somehow did not not consider an idea of having arpeggio as a `StaffObj`. I had an impression that the staff object status is reserved for elements that constitute the main content of a score and can have a meaning more or less independently of whether there are other objects nearby in the score. Objects that are always attached to other object or always connect something in a score seem to fall into other categories in the internal model so I have only considered choosing an object type for arpeggio between them. If the scheme with `StaffObj`s turns out to work better I certainly can rework it that way, but if having arpeggio as a `RelObj` is fine then it would indeed be better to consider this again when implementing a cross-staff (and cross-chord) arpeggio support.

cecilios:
Please, a question: I'm adding a couple of scores to the Lomse regression test set (I will commit them soon). The one from Lilypond is already in my set, so I replicated the examples in your post (“Arpeggio of different lengths” and “Arpeggio on cross-staff chords”). I've noticed differences in rendering between your image and what I get. All of them but one are probably due to resolution, scaling factors and document sizes, so they look fine to me. But there is one that I cannot explain: the vertical position of words “Andantino” and “dolce”; also the horizontal position of word “dolce” is different. All other texts are placed at the same positions in both scores but not these two. Can you find an explanation for this?

My image:
image

Your image: 
image


Also,  a request: please for future PRs I will appreciate if you could also include some unit tests. For engraving and layout I know that normally it is difficult to code tests; thus, the automated visual regression tests can be used instead for detecting bugs and changes in layout behaviour and I would not expect you to provide unit tests here. But I will appreciate unit tests for other library parts, such as the MusicXML importer, as they should be simple to code and are of great help to detect bugs when refactoring or adding more features. Thank you!

dmitrio95:
The difference is probably in `default-y` attributes of those text elements.  I have prepared based on the MusicXML code found [here](https://www.w3.org/2021/06/musicxml40/musicxml-reference/examples/tutorial-apres-un-reve), and it contains `default-y="15"` attributes for these texts. If I remove the `default-y` attributes these texts are positioned in the same way as it is shown on your image. Here is the file I have used to prepare my screenshot:
[musicxml_example_arpeggiated.musicxml.txt](https://github.com/lenmus/lomse/files/6933922/musicxml_example_arpeggiated.musicxml.txt)

As for unit tests, sure, I will add them in my future pull requests.

cecilios:
Confirmed. The difference is in default-x and default-y attributes. 
Thank you!

cecilios:
@dmitrio95 If you don't have plans in a near future to add support for arpeggiated chords in playback, I have time and can take this throughout this month. Please confirm me. Thank you.

dmitrio95:
Yes, I have no plans to work on playback issues in the nearest future. My current focus with Lomse is on improving engraving, adding various notation support and fixing crashes I have encountered with some MusicXML scores. If that helps I can file issues on things I plan to do but anyway I will unlikely get to anything related to playback in the nearest future.

cecilios:
Ok, thank you. I will take the task of arpeggios playback. I would like to establish a simple procedure so that we are informed of what each one does, to avoid that, by chance, both of us get to work on the same thing. For me it is enough with a private email informing that something is going to be addressed and, in any case, without any commitment to finalize it. Opening an issue and adding a comment "taken by xxxxxxx" is also a good way of doing it, as you prefer. The important thing is to avoid duplicate work. Thanks!

dmitrio95:
Well, I am fine with any of the options. Some of the issues I eventually work on are pretty small and are unlikely to overlap with anything but for larger tasks some sort of synchronization makes sense. I can send you a brief overview of the next things I was planning to do or open issues for them, whichever you prefer. Probably the nearest task for me would be finalizing pedal import and renedring — I have pedals half-working for a long time already but I still need to figure out a better way to support both pedal signs and lines and synchronize positions of both these types of pedals.

cecilios:
> Some of the issues I eventually work on are pretty small and are unlikely to overlap with anything but for larger tasks some sort of synchronization makes sense.

Yes, of course, the need is only for larger tasks. Then I will always open an issue (if not already in the list) and assign it to me. As for things that you would like to do, proceed as you prefer at every moment, send me a brief comment by PM or open an issue and self-assign it. I wouldn't like to create administrative burden, just to share information to avoid repeating work. And, as said, in no case the communication will be taken as a commitment to do it.

As to piano pedal marks, is you need help or would like to comment alternatives or ideas, feel free to ask. You can use issue #218 for this or PM, as you prefer.

dmitrio95:
Thanks, then I'll keep you informed on the issues I will take.

PR #303, opened 2021-08-08
MusicXML: fix adding spurious staves on reading mid-score <attributes> with <staves> tag

dmitrio95:
When Lomse reads a mid-score `<attributes>` tags which contains some value for `<staves>` larger than 1, it may attempt to add extra staves to the existing instrument, which may lead to adding spurious staves or even to a crash. This may occur with the scores exported from MuseScore if they contain section breaks, for example:
 - [lomse_attr_staves_crash.musicxml.txt](https://github.com/lenmus/lomse/files/6950064/lomse_attr_staves_crash.musicxml.txt) — crashes Lomse on engraving stage
(the original score in the MuseScore format: [lomse_attr_staves_crash.mscx.txt](https://github.com/lenmus/lomse/files/6950071/lomse_attr_staves_crash.mscx.txt))
 - [lomse_attr_staves_no_crash.musicxml.txt](https://github.com/lenmus/lomse/files/6950065/lomse_attr_staves_no_crash.musicxml.txt) — doesn't crash Lomse but displays a spurious third staff which shouldn't be there.

The cause of the issue is that the MusicXML importer assumes that the instrument has only one staff prior to reading `<attributes>` tag and adds staves on top of that regardless of whether they are needed. This PR corrects this behavior so no unnecessary staves are added. I am not sure whether something more sensible needs to be done if number of staves actually changes but if it doesn't this solution prevents the issues related to that.

### DIFF
--- a/src/parser/mxl/lomse_mxl_analyser.cpp
+++ b/src/parser/mxl/lomse_mxl_analyser.cpp
@@ -1994,10 +1994,11 @@ public:
         // staves?
         if (get_optional("staves"))
         {
-            int staves = get_child_value_integer(1);
+            const int targetStaves = get_child_value_integer(1);
             ImoInstrument* pInstr = dynamic_cast<ImoInstrument*>(m_pAnchor->get_parent_imo());
+            int instrStaves = pInstr->get_num_staves();
             // coverity[tainted_data]
-            for(; staves > 1; --staves)
+            for (; instrStaves < targetStaves; ++instrStaves)
                 pInstr->add_staff();
         }
 
--- a/src/tests/lomse_test_mxl_analyser.cpp
+++ b/src/tests/lomse_test_mxl_analyser.cpp
@@ -1353,6 +1353,81 @@ SUITE(MxlAnalyserTest)
         delete pRoot;
     }
 
+    TEST_FIXTURE(MxlAnalyserTestFixture, MxlAnalyser_attributes_04)
+    {
+        //@04 setting number of staves
+        stringstream errormsg;
+        Document doc(m_libraryScope);
+        XmlParser parser;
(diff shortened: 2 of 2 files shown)


cecilios:
Thank you by this PR. I will review it as soon as possible!

cecilios:
> I am not sure whether something more sensible needs to be done if number of staves actually changes but if it doesn't this solution prevents the issues related to that.

Currently the staves model is very simple: the number of staves for each instrument is defined in ImoInstrument but it is assumed that it never changes. Staves are modeled as a list of ImoStaffInfo objects, each one containing the information (spacing, number of lines, etc.) for one staff.

I have never tried to address adding support for scores in which the number of staves could be dynamically changed, increasing or decreasing it. For this, probably, the solution would require to remove staves definition from ImoInstrument and move this to an AuxObj attached to an ImoDirection or other ImoStaffObj, something similar to MusicXML `<attributes>` element.

I don't know how Lomse will currently behave if the number of staves is changed in the middle. If the number of staves is increased I don't foresee any crashes: it will behave as if the instrument had the maximum number of staves defined since beginning, and will be rendered with that maximum number since beginning but with empty staves in parts having less staves. But if the number is decreased, probably there will be crashes due to loops using the maximum defined number of staves instead of the current part number of staves. I never tested.

As to your PR, I think it is correct and nothing else should be done until support for dynamically changing the number of staves is added.

Perhaps this is one of the next important changes to introduce in Lomse, as scores in which the number of staves changes dynamically are common, e.g. ossia staves, although this can be a special case with a different treatment.

Thanks again for this PR.

dmitrio95:
Thanks for your comment! Then it is indeed better to postpone further changes until support for variable staves number is added.

PR #311, opened 2021-08-22
Fix a crash when a tuplet starts or ends with an invisible note

dmitrio95:
Invisible notes are quite often found in digital score formats to achieve certain playback effects. Lomse correctly imports `print-object` attribute of such notes and creates invisible shapes in place of them but this doesn't play well with engraving some relation objects which happen to include such invisible notes. Therefore I have encountered a number of crashes on engraving scores which include invisible notes.

This pull requests fixes a crash in case an invisible note is a start or end note in a tuplet (a small example score: [lomse_tuplet_invisible_note_crash.musicxml.txt](https://github.com/lenmus/lomse/files/7027305/lomse_tuplet_invisible_note_crash.musicxml.txt)). In this case it is enough just to rearrange a bit a logic of the tuplet engraver to explicitly check all assumptions about shape types which the engraver makes. Therefore the engraver's logic remains unchanged for the case when all notes and rests in the tuplet are visible but it now correctly handles the case when some of the tuplet's notes are invisible.

### DIFF
--- a/include/lomse_tuplet_engraver.h
+++ b/include/lomse_tuplet_engraver.h
@@ -102,7 +102,7 @@ protected:
     ImoNoteRest* get_start_noterest() { return m_noteRests.front().first; }
     ImoNoteRest* get_end_noterest() { return m_noteRests.back().first; }
 
-    void compute_y_coordinates();
+    void compute_y_coordinates(GmoShapeNote* pStart, GmoShapeNote* pEnd);
     GmoShapeNote* get_first_note();
     GmoShapeNote* get_last_note();
 
--- a/src/graphic_model/engravers/lomse_tuplet_engraver.cpp
+++ b/src/graphic_model/engravers/lomse_tuplet_engraver.cpp
@@ -131,6 +131,12 @@ GmoShape* TupletEngraver::create_first_or_intermediate_shape(LUnits UNUSED(xStaf
 //---------------------------------------------------------------------------------------
 GmoShape* TupletEngraver::create_last_shape(Color color)
 {
+    GmoShapeNote* pStart = get_first_note();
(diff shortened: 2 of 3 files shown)


cecilios:
I always had doubts about the convenience/need of importing invisible notes. They are just a trick to implement playback notations not supported by a player, e.g. a trill. But if the player supports these notations there will be a conflict in playback: the player will properly implement playback but invisible notes will duplicate or interfere with the playback.

With invisible rests the situation is different as they can be included to represent a gap in a voice, and do not affect playback.

Lomse objective for playback is to properly support all notation in playback. Thus, the first time I encountered invisible notes (very recently) I was considering the convenience of importing invisible notes/rests included in the file. But as invisible rest must be imported I decided to import invisible notes/rest and to avoid conflicts in playback by ignoring them in playback.

After locally merging this PR and doing some tests, I've noticed that the test score you attached to the PR renders the tuplet bracket for the invisible notes at top of page (probably the engraver does not have references as the shape is invisible):


pr311


Therefore, it is necessary to ensure that any additional notation attached to invisible notes is not rendered (or even not imported). 

As this PR certainly avoids crashes and I have not detected issues in regression tests, if you agree I will merge it without waiting for fixes to this issue of rendering attached notations. Nevertheless, as you prefer.

dmitrio95:
Ah, I have only had a picture of the visible part of the page so didn't notice that tuplet. It is better to get that fixed indeed. It is easy to ensure that a tuplet with no visible notes is not rendered at all by moving the check on start and end notes presence a bit earlier, if this change is fine. I see that the tuplet engraver doesn't seem to take rests into account but this fix doesn't change this situation anyway. I have updated the pull request with this change.

cecilios:
Thank you very much!

PR #313, opened 2021-08-29
Ensure that non-finished relation objects are engraved in the correct order

dmitrio95:
In order to ensure that different objects are engraved in the expected order Lomse makes several passes across the system's list of pending objects and engraves object of one type in each pass. However this doesn't work for objects that are not finished in the current system as those are engraved after engraving all objects that are finished in the current system. This leads to an inconsistent engraving order if unfinished relation objects are involved, for example (note that the order for a wedge and an octave shift is inverted when the octave shift spans for two systems):

This pull request moves engraving of non-finished relation objects to the same loop that does engraving of other objects which makes the engraving order more consistent between those objects:

(test file: [engraving-order-example.musicxml.txt](https://github.com/lenmus/lomse/files/7073072/engraving-order-example.musicxml.txt))

Of course, the situation shown in the example is not really common, but having a consistent engraving order will greatly simplify things in future. This will ensure the correct engraving order if more object types are added (for example, to ensure the engraving order between wedges and pedal lines which are typically both placed below a staff) and will make it easier to introduce vertical alignment of certain element types (for example, consecutive volta brackets, pedals, combinations of dynamic marks and hairpins etc.)

### DIFF
--- a/src/graphic_model/layouters/lomse_system_layouter.cpp
+++ b/src/graphic_model/layouters/lomse_system_layouter.cpp
@@ -796,22 +796,26 @@ void SystemLayouter::engrave_system_details(int iSystem)
                     ++it;
             }
         }
-    }
 
-    //engrave RelObjs that continue in next system
-    list<PendingRelObj>::iterator itR;
-    for (itR = m_pScoreLyt->m_notFinishedRelObj.begin();
-         itR != m_pScoreLyt->m_notFinishedRelObj.end(); ++itR)
-    {
-        engrave_not_finished_relobj((*itR).first, (*itR).second, iSystem);
-    }
-
-    //engrave Lyrics that continue in next system
-    list<PendingLyricsObj>::iterator itAR;
-    for (itAR = m_pScoreLyt->m_notFinishedLyrics.begin();
-         itAR != m_pScoreLyt->m_notFinishedLyrics.end(); ++itAR)
-    {
-        engrave_not_finished_lyrics((*itAR).first, (*itAR).second, iSystem);
+        if (k_imo_relobj < type && type < k_imo_relobj_last)
+        {
+            //engrave RelObjs that continue in next system
+            for (PendingRelObj& obj : m_pScoreLyt->m_notFinishedRelObj)
+            {
+                if (obj.first->get_obj_type() == type)
+                {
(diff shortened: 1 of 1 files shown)


cecilios:
Thank you very much for this PR.

> but having a consistent engraving order will greatly simplify things in future.

Totally agree. This is an important change that should have been done time ago! Thank you.

cecilios:
While testing PR#314 I've noticed a regression error  (e.g lilypond test [31a-Directions.xml.txt](https://github.com/lenmus/lomse/files/7076347/31a-Directions.xml.txt) ) caused by this PR that I overlooked when testing this PR. Now, after merging this PR,  wedges, volta brackets and 8va signs are placed below lyrics lines.

It is necessary to change priorities in the engraving priority list as I've seen that new elements supported after the list was created have been added at the end without more thinking as they didn't interfere. I will fix this.

cecilios:
Fixed in https://github.com/lenmus/lomse/commit/909763358da623faed900ff63a24c4c457146150

dmitrio95:
Thanks, I have indeed not reviewed the engraving order here. In addition to the changes in 909763358da623faed900ff63a24c4c457146150, I think another logical change can be placing dynamic marks and wedges to adjacent positions in the engraving order table (not sure where exactly though). They both refer to dynamics and should be placed vertically close to each other or can even be vertically aligned in some cases. This would be sometimes problematic if other objects are engraved between the engraving of dynamic marks and wedges.

I can do it in my next pull request, where I plan to propose an implementation of vertical alignment of some auxiliary and relation objects (as I have briefly mentioned, not only dynamics and wedges but also voltas and later pedals, perhaps also something like [`dashes`](https://w3c.github.io/musicxml/musicxml-reference/elements/dashes/) object and its corresponding text). This alignment would only make sense in conjunction with a consistent engraving order for these types.

cecilios:
> I have indeed not reviewed the engraving order here.

I neither. I'm still on travel and could not check engraving order in notation books. So I just moved some notations before the lyrics line. Please feel free to change the order as you think it is better.

> I can do it in my next pull request, where I plan to propose an implementation of vertical alignment of some auxiliary and relation objects ...

That will be a great improvement! Thank you

PR #317, opened 2021-09-04
Implement relative alignment of auxiliary objects, align voltas and hairpin-dynamic sequences

dmitrio95:
This pull request implements a mechanism to align auxiliary object with respect to each other, both horizontally and vertically. The implementation proposed here involves an introduction of a "horizontal layer" concept which basically means a set of objects which should be aligned with respect to each other if they are placed close enough. This concept is implemented by the `HorizontalLayer` and related classes, also the engraving order table was restructured to include the information about horizontal layers. There is one layer per placement per staff for each system (that is, shapes placed below and above staff are aligned separately). The information on shape placement doesn't seem to be readily available at the `SystemLayouter` level, so it is currently a responsibility of engravers to put their shapes to relevant horizontal layers.

Horizontal layers provide an information about objects already added to them (which is useful for engravers) and align sequences of shapes which are close to each other by a common baseline.

Currently engravers can access horizontal layers via `VerticalProfile` which is something I am not really sure about. On the one side these are separate concepts, but on the other side they both serve the purpose of improving placement of shapes with respect to other score objects, and it seems that every engraver which would need a `HorizontalLayer` would also really need a `VerticalProfile` as well. So for now I have included a pointer to the current horizontal layer system into `VerticalProfile` class to avoid explicitly passing another argument to engravers. This can certainly be changed though.

This alignment was applied to volta brackets and to wedge-dynamic combinations. Wedges are aligned both horizontally and vertically with other shapes so if a dynamic mark is placed at the same position where a wedge starts the wedge is no longer pushed away from the staff:
satie

The alignment effect is limited by a distance of the aligned objects (for volta brackets the limit is infinite), so something like this will not be forcefully aligned by the horizontal layer:
noalign

Also MusicXML scores may contain dynamic marks which are invisible in the original score but are not marked so in the MusicXML file due to a lack of `print-object` attribute for `direction` objects. In this PR newer shapes added to a horizontal layer exclude the existing overlapping shapes from the further alignment algorithm which allows to handle such situations more or less gracefully:
invisible-dynamics

In addition to the relative object alignment this PR also includes the following changes:
 - Moves dynamic marks right before wedges in the engraving order. This is necessary to be able to properly align them as described above.
 - Adds shifting a dynamic mark shape with respect to the system's vertical profile. This is not exactly necessary for anything else in this pull request, still it helps to avoid collisions of dynamic marks with beams.

### DIFF
--- a/add-sources.cmake
+++ b/add-sources.cmake
@@ -107,6 +107,7 @@ set(GRAPHIC_MODEL_FILES
     ${LOMSE_SRC_DIR}/graphic_model/engravers/lomse_volta_engraver.cpp
     ${LOMSE_SRC_DIR}/graphic_model/engravers/lomse_wedge_engraver.cpp
 
+    ${LOMSE_SRC_DIR}/graphic_model/layouters/lomse_aux_shapes_aligner.cpp
     ${LOMSE_SRC_DIR}/graphic_model/layouters/lomse_blocks_container_layouter.cpp
     ${LOMSE_SRC_DIR}/graphic_model/layouters/lomse_document_layouter.cpp
     ${LOMSE_SRC_DIR}/graphic_model/layouters/lomse_inlines_container_layouter.cpp
--- a/include/lomse_aux_shapes_aligner.h
+++ b/include/lomse_aux_shapes_aligner.h
@@ -1,6 +1,6 @@
 //---------------------------------------------------------------------------------------
 // This file is part of the Lomse library.
-// Lomse is copyrighted work (c) 2010-2016. All rights reserved.
+// Lomse is copyrighted work (c) 2010-2021. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without modification,
 // are permitted provided that the following conditions are met:
@@ -27,45 +27,61 @@
 // the project at cecilios@users.sourceforge.net
 //---------------------------------------------------------------------------------------
 
-#include "lomse_engraver.h"
+#ifndef __LOMSE_AUX_SHAPES_ALIGNER_H__
+#define __LOMSE_AUX_SHAPES_ALIGNER_H__
 
-#include "lomse_internal_model.h"
-#include "lomse_aux_shapes_aligner.h"
-#include "lomse_score_meter.h"
-#include "lomse_vertical_profile.h"
+#include "lomse_basic.h"
 
+#include <map>
+#include <vector>
 
 namespace lomse
 {
 
-//---------------------------------------------------------------------------------------
-// Engraver implementation
-//---------------------------------------------------------------------------------------
-LUnits Engraver::tenths_to_logical(Tenths value) const
+class GmoShape;
+class VerticalProfile;
+
+class AuxShapesAligner
 {
-    return m_pMeter->tenths_to_logical(value, m_iInstr, m_iStaff);
-}
+protected:
+    std::map<LUnits, GmoShape*> m_shapes;
+    LUnits m_xAbsLeft;
+    LUnits m_xAbsRight;
 
-double Engraver::determine_font_size()
+    using ShapeIterator = std::map<LUnits, GmoShape*>::const_iterator;
+
+public:
+    AuxShapesAligner(LUnits xAbsLeft, LUnits xAbsRight);
+
+    void add_shape(GmoShape* pShape);
+    void align_shape_base_lines(LUnits maxAlignDistance, VerticalProfile* pVProfile, int idxStaff, bool fDown);
+
+    GmoShape* find_shape(LUnits x) const;
+    LUnits find_nearest_free_point_left(LUnits x) const;
+    LUnits find_nearest_free_point_right(LUnits x) const;
+    LUnits find_nearest_occupied_point_left(LUnits x) const;
+    LUnits find_nearest_occupied_point_right(LUnits x) const;
+
+protected:
+    void set_base_line_for_range(ShapeIterator rangeBegin, ShapeIterator rangeEnd,
+                                 LUnits yBaseline, VerticalProfile* pVProfile, int idxStaff);
+};
+
+class AuxShapesAlignersSystem
 {
-    return 21.0 * m_pMeter->line_spacing_for_instr_staff(m_iInstr, m_iStaff) / 180.0;
-}
+protected:
+    std::vector<AuxShapesAligner> m_alignersAbove;
+    std::vector<AuxShapesAligner> m_alignersBelow;
+    LUnits m_maxAlignDistance;
 
-void Engraver::add_user_shift(ImoContentObj* pImo, UPoint* pos)
-{
-    (*pos).x += tenths_to_logical(pImo->get_user_location_x());
-    (*pos).y += tenths_to_logical(pImo->get_user_location_y());
-}
+public:
+    AuxShapesAlignersSystem(size_t numStaves, LUnits xAbsLeft, LUnits xAbsRight,
+                          LUnits maxAlignDistance);
 
-//---------------------------------------------------------------------------------------
-// RelObjEngraver implementation
-//---------------------------------------------------------------------------------------
-void RelObjEngraver::add_to_aux_shapes_aligner(GmoShape* pShape, bool fAboveStaff)
-{
-    AuxShapesAligner* pAligner = m_pVProfile->get_current_aux_shapes_aligner(m_idxStaff, fAboveStaff);
-    if (pAligner)
-        pAligner->add_shape(pShape);
-}
+    AuxShapesAligner& get_aligner(int staff, bool fAbove) { return fAbove ? m_alignersAbove[staff] : m_alignersBelow[staff]; }
+    void align_shape_base_lines(VerticalProfile* pVProfile);
+};
 
+}   //namespace lomse
 
-}  //namespace lomse
+#endif      //__LOMSE_AUX_SHAPES_ALIGNER_H__
--- a/include/lomse_dynamics_mark_engraver.h
+++ b/include/lomse_dynamics_mark_engraver.h
@@ -42,6 +42,7 @@ class ImoDynamicsMark;
 class GmoShape;
 class ScoreMeter;
 class GmoShapeDynamicsMark;
+class VerticalProfile;
 
 //---------------------------------------------------------------------------------------
 class DynamicsMarkEngraver : public Engraver
@@ -52,10 +53,12 @@ protected:
     bool m_fAbove;
     GmoShape* m_pParentShape;
     GmoShapeDynamicsMark* m_pDynamicsMarkShape;
+    int m_idxStaff;
+    VerticalProfile* m_pVProfile;
 
 public:
     DynamicsMarkEngraver(LibraryScope& libraryScope, ScoreMeter* pScoreMeter,
-                    int iInstr, int iStaff);
+                    int iInstr, int iStaff, int idxStaff, VerticalProfile* pVProfile);
     ~DynamicsMarkEngraver() {}
 
     GmoShapeDynamicsMark* create_shape(ImoDynamicsMark* pDynamicsMark, UPoint pos,
@@ -68,6 +71,7 @@ protected:
     void center_on_parent();
     void add_voice();
     int find_glyph();
+    void shift_shape_if_collision();
 
 };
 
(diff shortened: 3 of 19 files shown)


cecilios:

include/lomse_horizontal_layer.h
```diff
@@ -0,0 +1,96 @@
...
+protected:
+    struct OccupiedSpaceInterval
+    {
+        GmoShape* pShape;
+        LUnits xLeft;
+        LUnits xRight;
+    };
+
+    std::map<LUnits, OccupiedSpaceInterval> m_intervals;
+    LUnits m_xAbsLeft;
```
cecilios, line 55: I don't see any problem here and there is no need to change anything. I'm just curious about why you designed it in this way. The OccupiedSpaceInterval struct contains the shape and the occupied interval, but the interval is always from shape->left() to shape->right(). Wouldn't it be simpler to just save the shape and define `m_intervals` as `std::map<LUnits, GmoShape*> m_intervals;` ? By using the struct I don't see much gain in speed and requires more memory. But as the number of objects to align in a system is very small, this is not a real problem. So I'm neither asking for changes nor questioning the design. I'm just asking because I would like to understand better your reasoning. Your implementation of HorizontalLayer using a map is very clever and clean, and probably there is some benefit in using the struct that I'm missing.

src/graphic_model/layouters/lomse_horizontal_layer.cpp
```diff
@@ -0,0 +1,234 @@
...
+    for (auto it = rangeBegin; it != rangeEnd; ++it)
+    {
+        GmoShape* pShape = it->second.pShape;
+        const LUnits yShift = yBaseline - pShape->get_baseline_y();
+
+        if (yShift)
+        {
+            pShape->shift_origin(USize(0, yShift));
+            pVProfile->update(pShape, idxStaff);
+        }
```
cecilios, line 131: A comment, for your information, just in case.

VerticalProfile was designed considering that once a shape is added to it, the shape position will not change. When a shape is added to VerticalProfile its boundaries are included in the profile. If the position of the shape is later modified, a new invocation to `VerticalProfile::update()` does not remove the old computed positions, just includes the new ones. This will imply that unoccupied space could remain marked as occupied.

But, in this method, the shift is always in vertical, up for the above HorizontalLayer and down for the bellow HorizontalLayer. Thus updating VerticalProfile does the right thing, and there is no problem. Perhaps VerticalProfile should be improved for supporting updates when a shape is moved when overlaps or by other reason (I take note).

cecilios:
My overall impression is that this HorizontalLayer tool is a great improvement, and its implementation, using a map, is very clever and clean, and that this PR has to be merged. Thank you!

Some general comments:

**1. Accessing the HorizontalLayer via VerticalProfile**

I don't like the idea of including the HorizontalLayer in VerticalProfile as they are not functionally related. But it is true than passing more arguments to engravers is not a good idea. Engravers are already receiving too many parameters and this has to be reviewed and fixed (I take note). A solution for accessing system level objects without passing a pointer to each needed object should be studied. Until then, I agree in accessing the HorizontalLayer via VerticalProfile as there are no unit tests for the involved classes.  So no changes in this PR.

**2. HorizontalLayer name**

I would propose to change the class name from HorizontalLayer to HorizontalAligner. To me ‘Layer’ suggests something placed over another object and this is not the idea of this class. To be coherent, it would require other changes such as variables named ‘layer’ to be renamed as ‘aligner’ and the header guard  `__LOMSE_HORIZONTAL_LAYER_H__`, but these changes are less important.  What is your opinion?  In any case, I can accept the HorizontalLayer name.

**3. Observed behavior**

Finally a comment about changes observed in visual regression tests. This PR certainly improves the layout: wedges and dynamic marks are now aligned, wedges width is now adjusted so that dynamic marks can be aligned before/after the wedge, and volta lines are now aligned. And certainly this HorizontalLayer mechanism could also be used to align pedal lines. 

There is one test, score [MahlFaGe4Sample.musicxml.txt](https://github.com/lenmus/lomse/files/7111489/MahlFaGe4Sample.musicxml.txt) from the MusicXML samples set, in which the alignment of wedges, although acceptable, will require future improvements in Lomse. In the original score the wedges in third measure are not aligned with the previous ones, probably for keeping more blank separation space with previous system:


image




But in Lomse rendition the system starts just after the second wedge. Now, after merging this PR, the second wedge is aligned with the third one. This is not incorrect but as the wedge is more near to previous system than to its system the layout could be confusing:


image




So the first change I would do is to add more separation between systems to avoid doubts about which system the notation refer to, specially in case of wedges. This has nothing to do with this PR and I'm not requesting any change. I just will add this to my notes, not to forget this for future. What is your opinion?

A second change I see, but not sure about its utility/convenience and would appreciate your point of view, is to limit wedge alignment (or other objects, in general) when it implies a big shift, as this could unnecessarily increase the height of a system if more notation is added on top of the shifted objects. So, nothing to change in this PR. Again, I just take note and will keep an eye open to confirm or not the utility/convenience of this change in future.

A great PR. Thank you!

include/lomse_horizontal_layer.h
dmitrio95: Yes, you are right, in the current version the structure is not needed. This is actually a leftover from my previous versions of `HorizontalLayer`: initially I thought it would be good to split or adjust boundaries of intervals if some of them overlap with the newly added shape, just to keep track on all shapes added to the layer. In this case interval boundaries could differ from the corresponding shapes' boundaries. However when I checked this with some real scores I realized that this creates a total mess when there are some horizontally overlapping shapes so I went for just removing overlapping shapes from the horizontal layer. Now interval boundaries indeed always equal to the boundaries of the corresponding shapes, and no special structure for an interval is needed. I think I'll remove the structure and replace it with a plain pointer to `GmoShape`, thanks for your comment!

src/graphic_model/layouters/lomse_horizontal_layer.cpp
dmitrio95: Yes, the shift here is always vertical and always away from a staff (at least if shapes are added to the correct layers). I don't actually think it is necessary now to support horizontal displacement in `VerticalProfile`, because horizontal placement of shapes is currently always decided before adding them to the vertical profile. If some change in the engraving algorithm requires that then this can be implemented along with that change.

dmitrio95:
> Engravers are already receiving too many parameters and this has to be reviewed and fixed (I take note).

For some parameters it would be enough to group them together if they are related. For example, some engraver parameters are already grouped into the `PendingAuxObj` structure but later they are unpacked and passed to engravers as separate arguments. Of course, `PendingAuxObj` seems to contain more information than engravers need so it should probably not be passed there directly but the approach of creating helper structures with related information may possibly help to improve the situation.

> I would propose to change the class name from HorizontalLayer to HorizontalAligner.

Yes, I agree that `HorizontalLayer` is probably not the best name as it overlaps with the concept of layers which are drawn on top of each other. This is indeed some sort of aligner entity but the `HorizontalAligner` name, in my opinion, would not be exactly correct as this name may imply that only horizontal alignment is handled by this class while in fact it also handles vertical alignment of objects added to it. Therefore perhaps some other related name could be chosen, like `AttachedShapeAligner`, `SystemShapeAligner` or `AuxShapeAligner`? I am not sure how it would be better to describe both `AuxObj`s and `RelObj`s in one word but perhaps some term like this could describe the purpose of this class better.

> So the first change I would do is to add more separation between systems to avoid doubts about which system the notation refer to, specially in case of wedges.

Do you mean that space between systems should always be increased or only in cases when there is some potentially confusing musical notation placement? In situations like in your example this indeed would help but in most cases I would say system distances are fine in Lomse. So perhaps something like enforcing a minimal distance between systems' vertical profiles could be useful?

> A second change I see, but not sure about its utility/convenience and would appreciate your point of view, is to limit wedge alignment (or other objects, in general) when it implies a big shift, as this could unnecessarily increase the height of a system if more notation is added on top of the shifted objects.

This would be not difficult to implement within the current pull request, it is just that the alignment algorithm in `HorizontalLayer` would need to check not only a horizontal distance between shapes to align but also a vertical distance between their baselines. The question is, whether only a vertical displacement should be the criterion or something different, like combination of both horizontal and vertical distance? For example, in [the first example](https://user-images.githubusercontent.com/6000747/132096926-1e947d04-bd83-428f-836b-8a1be3f10a93.png) from my initial post the dynamic mark is vertically displaced to a similarly large distance but in that case that displacement seems to be the right thing to do.

dmitrio95:
I have updated the pull request to use `GmoShape*` pointers instead of special structures in the `HorizontalLayer` implementation. I will also rename the `HorizontalLayer` class when we decide on a better name for it.

cecilios:
> Do you mean that space between systems should always be increased or only in cases when there is some potentially confusing musical notation placement?

No, only when the distance is minimal. Currently in https://github.com/lenmus/lomse/blob/58d475edfbd8f52c0f3c4804137accc0ac8db321/src/graphic_model/layouters/lomse_system_layouter.cpp#L681-L698 it is checked if systems overlap and in that case a minimum distance is enforced. But if they do not overlap, this minimal distance is not enforced. So, as you suggests, a first step is to enforce always a minimal distance.

> The question is, whether only a vertical displacement should be the criterion or something different,..

Yes, totally agree. I'm not sure if this is needed so I would not do any change now. I only liked to raise the issue so that we pay attention to situations like this one in new scores, to collect evidence and information and then deciding if it is necessary to do anything or not, And collect the problematic examples found, if any. Your idea of using a combination of both horizontal and vertical distance seems a better starting point, in case we conclude that it is necessary to do something.

> PendingAuxObj seems to contain more information than engravers need so it should probably not be passed there directly

I don't see a problem in passing structures containing more information, as long as these structures are simple to create for unit tests and can be used by most engravers. Nevertheless, I agree that the solution would be to create helper structures, possible enclosing all 'global' system information that could be needed by engravers, but again considering the cost of instantiating these structures for unit tests. It is true that for engravers and layout it is complex to do unit tests and although at first I tried to create these tests, nowadays for the layout/engraving code I only create unit test in very specific situations and most of the code is tested via visual regression test, by using specific tests scores created for testing the different situations.

> the HorizontalAligner name, in my opinion, would not be exactly correct

It is difficult to find short names transmitting the full responsibility of one object, so for me it is enough if the name transmits the main idea. Thus, I don't see a bad choice in using the word 'Horizontal' as the idea of horizontal alignment somehow implies that objects have to be vertically moved so that they share a common horizontal baseline. For me the problem is in the word 'Layer'. But if you like the idea of changing the name, any name transmitting the idea of alignment (XxxxAligner) or position (XxxxPositioner) would be good for me. My preference, from the names you suggest, is AuxShapeAligner, but in plural, AuxShapesAligner. But use the name you prefer.


> it would be better to describe both AuxObjs and RelObjs in one word

Totally agree. This is a recurrent problem I find many times when I try to refer to both. In the classes three there is no a common ancestor at the same level than StaffObj:
```
+-- ImoScoreObj
      |
      +-- ImoStaffObj
      |
      +-- ImoAuxObj- single attachment, not related to other attachments
      | 
      +-- ImoRelObj - Relates two or more StaffObjs
```
So it is difficult to find a name for them. Both are "auxiliary objects" but AuxObj refers only to a subgroup. If you have ideas for a name to refer both AuxObjs and RelObjs,  I'll be glad to know as this issue of a name for referring to both is recurrent.

include/lomse_horizontal_layer.h
cecilios: Understood. Thank you

src/graphic_model/layouters/lomse_horizontal_layer.cpp
cecilios: Thank you. Hope this is not necessary as in that case the VerticalProfile class will have to be totally  redesigned!

dmitrio95:
OK, I have renamed the class to `AuxShapesAligner` and adjusted other related names.

> If you have ideas for a name to refer both AuxObjs and RelObjs

The only idea which currently comes to my mind is perhaps something like `AttachedObj` (or "attached object" in plain English) which would denote the fact that these objects are always attached to other objects and are not used independently. Still it may be not the best term for this, just an idea.

cecilios:
Merged. Thank you!

AttachedObj is probably a very good name as AuxObjs and RelObjs are "attached objects". I'm going to think on this and tell you something. Thank you!

PR #318, opened 2021-09-13
Fix overlaps when two voices in the same staff

cecilios:
This PR fixes layout errors due to overlaps in notes, chords and rests when there are two voices in the same staff (e.g. test score 00633):

Lomse (before this PR):
image

Lomse (after this PR):
image

The solution has been to create a NoterestsCollisionsFixer object that will receive all note/rest shapes at a given staff and timepos, to check for overlaps and solve any detected issue by moving sideways the conflicting objects. The NoterestsCollisionsFixer object is created in the spacing algorithm in the TimeSliceNoterest objects.

When this task was started, the collision scenarios I had in mind were very simple, just unisons, notes a second apart and rests and notes collisions. But as I started to work and studied collision scenarios and engraving rules in these situations, it resulted that there are many engraving rules involved and the number of collision scenarios is overwhelming, requiring in practice near a case by case solution. 

Therefore, this PR is a first step, fixing most collision cases between chords, notes, rests and accidentals. But not cases involving dots and collisions with flags, that introduce an extra level of complexity. So I prefer to take a rest on this issue for a while.

**Limitations**

Current implementation has some limitations:

- The solution is currently restricted to fix overlaps in two voices. If there are more than two voices in the same staff the collisions will be fixed for two voices but nothing will ensure that there are no overlaps with the other voices.

- Overlaps between noteheads one third apart are considered false positives as both noteheads should touch each other. But when the noteheads are of different types (e.g. a quarter note and a half note) it can be a real overlap and this is not fixed in this PR.

I preferred not to deal with these two cases until having real scores displaying these types of problems, to confirm the need and to have more knowledge about the specific problems and the possible solutions.

Also, as commented, this PR does not deal with dots and with cases involving collisions with flags.


src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp
```diff
@@ -0,0 +1,1149 @@
...
+}
+
+//---------------------------------------------------------------------------------------
+void NoterestsCollisionsFixer::add_noterest(GmoShape* pShape, ColStaffObjsEntry* pEntry)
+{
+//    LOMSE_ASSERT(pEntry->imo_object()->is_note_rest());
+//    LOMSE_ASSERT(pShape->is_shape_note() || pShape->is_shape_rest());
+
+    ImoNoteRest* pNR = static_cast<ImoNoteRest*>(pEntry->imo_object());
+    int posOnStaff = (pNR->is_note() ? static_cast<GmoShapeNote*>(pShape)->get_pos_on_staff()
```
dmitrio95, line 83: This may be something similar to what I have addressed in #311: if an internal object is a note its shape is not necessarily a `GmoShapeNote`. Perhaps it would be better to check the type for the shape itself (or perhaps also exclude invisible notes from this algorithm)?

include/lomse_gm_basic.h
```diff
@@ -288,7 +288,7 @@ class GmoObj
     void set_left(LUnits xLeft);
     void set_top(LUnits yTop);
     virtual void shift_origin(const USize& shift);
-    void shift_origin(LUnits x, LUnits y);
+    virtual void shift_origin(LUnits x, LUnits y);
```
dmitrio95, line 291: Another possible option I see is to leave this function non-virtual and just make it call `shift_origin(USize(x, y))`. Then implementations of these two functions will always behave the same way, without the need to override both of them.

src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp
cecilios: Tanks. Very good point. I never think about invisible shapes as I still don't assume that there are invisible notes and rests in MusicXML; it is an idea that I do not finish assimilating. I'm contrary to using invisible objects as, to me, it is a trick to solve some deficiency in the model. 

I will review this PR for ensuring invisible notes are excluded.

include/lomse_gm_basic.h
cecilios: Thank you. You are right, current code duplicates the implementation with the risk of introducing different behavior. I will change this.

cecilios: Finally, it has to be defined as virtual. Otherwise, if the `GmoCompositeShape` is managed as a pointer to a shape (`GmoShape`), then the base class method is invoked instead of the shadowing method, which is logical. So if method with two parameters is used, i has to be defined as virtual. I left the override, but reviewed implementations to avoid repeating the code.

dmitrio95: We don't need to add shadowing methods if the method in the base class is defined by calling the virtual method, it is enough just to override the virtual method in inherited classes. My proposal is just to remove `shift_origin(LUnits x, LUnits y)` from all classes except the base class (`GmoObj` in this case) and re-define only `shift_origin(const USize&)` method in inherited classes. Then if the first method is defined in the base class like this:
```
void shift_origin(LUnits x, LUnits y) { shift_origin(USize(x, y)); }
```
then the virtual `shift_origin` flavor will always be called, always with the implementation suitable for the corresponding object type.

cecilios: Thank you. The approach you suggests is correct. I've tried but after several test I'm not able to make my code work. I don't know were is the problem, a bug in my compiler `g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0` or a definition problem in my code, but if I only define in base class `GmoObj`:
```
void shift_origin(LUnits x, LUnits y) { shift_origin(USize(x, y)); }
```
the compiler does not recognizes this method as an inherited method in GmoShapeRest class (derived from GmoCompositeShape) and I get two errors:

```
 /datos/cecilio/lm/projects/lomseapp/lomse/trunk/src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp: In member function ‘void lomse::NoterestsCollisionsFixer::move_rest(size_t, lomse::LUnits)’:
/datos/cecilio/lm/projects/lomseapp/lomse/trunk/src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp:628:42: error: no matching function for call to ‘lomse::GmoShapeRest::shift_origin(float, lomse::LUnits&)’
     pRestShape->shift_origin(0.0f, yShift);
                                          ^
In file included from ../../lomse/trunk/include/private/lomse_internal_model_p.h:52:0,
                 from ../../lomse/trunk/include/lomse_internal_model_defs.h:35,
                 from ../../lomse/trunk/include/lomse_internal_model.h:33,
                 from /datos/cecilio/lm/projects/lomseapp/lomse/trunk/src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp:33:
../../lomse/trunk/include/lomse_shape_base.h:156:18: note: candidate: virtual void lomse::GmoCompositeShape::shift_origin(const USize&)
     virtual void shift_origin(const USize& shift) override;
                  ^~~~~~~~~~~~
../../lomse/trunk/include/lomse_shape_base.h:156:18: note:   candidate expects 1 argument, 2 provided
/datos/cecilio/lm/projects/lomseapp/lomse/trunk/src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp: In member function ‘void lomse::NoterestsCollisionsFixer::move_notehead(lomse::GmoShapeNote*, lomse::LUnits)’:
/datos/cecilio/lm/projects/lomseapp/lomse/trunk/src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp:1001:46: error: no matching function for call to ‘lomse::GmoShapeNote::shift_origin(lomse::LUnits&, float)’
         pShapeNote->shift_origin(xShift, 0.0f); //shift all to right
                                              ^
In file included from ../../lomse/trunk/include/private/lomse_internal_model_p.h:52:0,
                 from ../../lomse/trunk/include/lomse_internal_model_defs.h:35,
                 from ../../lomse/trunk/include/lomse_internal_model.h:33,
                 from /datos/cecilio/lm/projects/lomseapp/lomse/trunk/src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp:33:
../../lomse/trunk/include/lomse_shape_base.h:156:18: note: candidate: virtual void lomse::GmoCompositeShape::shift_origin(const USize&)
     virtual void shift_origin(const USize& shift) override;
                  ^~~~~~~~~~~~
../../lomse/trunk/include/lomse_shape_base.h:156:18: note:   candidate expects 1 argument, 2 provided

```

So after several tests and changes without solving the issue, as the method with two parameters is only used in tree or four places and, at the end, in base class, the USize object will be created to invoke the method doing the work, I've opted for not lossing more time and definitly removing the method with to arguments.

dmitrio95: Yes, if one of those `shift_origin`s can be removed that would be even simpler.

Otherwise this PR looks good for me, thanks a lot!

cecilios:
Thank you. Merged

PR #322, opened 2021-09-21
MusicXML: Import <defaults> element

cecilios:
This PR imports the `<defaults>` element, with the exception of the `<concert-score>` element, useless for now, and the `<appearance>` values, also useless for now and because importing them will require to make engraving options not constant and this is not an urgent change.

The main achievements with this PR are:
- Page size and margins are now imported and taken into account. 
- System and staves distance and margins are also imported and taken into account. But, for distances (not for margins), the imported values could be overridden by the layout algorithm if more space was required.
- Font for words is now imported and used for texts.
- Font for lyrics and language for lyrics is now imported and used for lyric lines.
- Font for music is also imported but it is useless as Lomse will, for now, continue using Bravura font.

Imported scores should now properly display lyric lines when using non-Latin languages. For instance, test score 
[Echigo-Jishi.musicxml.txt](https://github.com/lenmus/lomse/files/7204803/Echigo-Jishi.musicxml.txt) from MusicXML test set:

Lomse, before this PR:

image

Lomse, after this PR:

image

This PR also:

- Adds method `ImoScore::tenths_to_logical(Tenths value)` to provide tenths to LUnits conversion when tenths not referred to an staff. This method will, perhaps in future, replace current method for conversion when no staff involved.
- Paper size and related methods has been removed from ImoScore. They were here by historical reasons but were not used.
- Lomse used a model with three values: left margin, right margin and binding margin. But MusicXML, instead of using a common binding margin, uses a model with four independent values. Also for top and bottom margins MusicXML allows to specify different values for odd and even pages. As MusicXML model is more flexible, and to fully support MusicXML, the Lomse internal model has been changed. Therefore, this PR introduces some backward incompatible changes in Document API: the methods for defining margins have been split into two groups: one for even pages and one for odd pages:
```
//Before this PR:
LUnits page_left_margin() const;
LUnits page_right_margin() const;
LUnits page_top_margin() const;
LUnits page_bottom_margin() const;
void set_page_left_margin(LUnits value);
void set_page_right_margin(LUnits value);
void set_page_top_margin(LUnits value);
void set_page_bottom_margin(LUnits value);

//After this PR:
    //odd pages
LUnits page_left_margin_odd() const;
LUnits page_right_margin_odd() const;
LUnits page_top_margin_odd() const;
LUnits page_bottom_margin_odd() const;
void set_page_left_margin_odd(LUnits value);
void set_page_right_margin_odd(LUnits value);
void set_page_top_margin_odd(LUnits value);
void set_page_bottom_margin_odd(LUnits value);
    //even pages
LUnits page_left_margin_even() const;
LUnits page_right_margin_even() const;
LUnits page_top_margin_even() const;
LUnits page_bottom_margin_even() const;
void set_page_left_margin_even(LUnits value);
void set_page_right_margin_even(LUnits value);
void set_page_top_margin_even(LUnits value);
void set_page_bottom_margin_even(LUnits value);
```

@dmitrio95 Probably you have some MusicXML samples with texts and lyrics in Russian language. If so and it is possible for you, I would appreciate if you could send me a couple of samples, for testing and reference, by PM if you prefer. Thank you!

### DIFF
--- a/include/lomse_document.h
+++ b/include/lomse_document.h
@@ -88,18 +88,29 @@ public:
     float page_content_scale() const;
     void set_page_content_scale(float scale);
 
-    //Document page size
-    LUnits page_left_margin() const;
-    LUnits page_right_margin() const;
-    LUnits page_top_margin() const;
-    LUnits page_bottom_margin() const;
+    //Document page size and margins
+        //odd pages
+    LUnits page_left_margin_odd() const;
+    LUnits page_right_margin_odd() const;
+    LUnits page_top_margin_odd() const;
+    LUnits page_bottom_margin_odd() const;
+    void set_page_left_margin_odd(LUnits value);
+    void set_page_right_margin_odd(LUnits value);
+    void set_page_top_margin_odd(LUnits value);
+    void set_page_bottom_margin_odd(LUnits value);
+        //even pages
+    LUnits page_left_margin_even() const;
+    LUnits page_right_margin_even() const;
+    LUnits page_top_margin_even() const;
+    LUnits page_bottom_margin_even() const;
+    void set_page_left_margin_even(LUnits value);
+    void set_page_right_margin_even(LUnits value);
+    void set_page_top_margin_even(LUnits value);
+    void set_page_bottom_margin_even(LUnits value);
+
     USize page_size() const;
     LUnits page_width() const;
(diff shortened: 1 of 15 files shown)


dmitrio95:
Sure, I have sent you some sample scores by e-mail.

PR #323, opened 2021-09-25
Cautionary time and key changes at end of system

cecilios:
This PR fixes the issue of adding a cautionary indication at the end of a system, when a change of key and/or time signature occurs after the system break. For instance, test score [11a-TimeSignatures.xml.txt](https://github.com/lenmus/lomse/files/7230019/11a-TimeSignatures.xml.txt) from Lylipond set:

Lomse, before this PR:
   
image

Lomse, after this PR:
   
image

Preparing this change has been simpler than I thought. The approach followed has been to move the system break after the new key and/or time signatures, when they exist, and to add a time signature to the next system prolog (in Lomse terminology the prolog is the clef, key and, optionally, time signature symbols placed at start of every system). Thus, there was no need to costly modifications in the spacing algorithm or other related classes.

**Changes**

- Method `ColumnBreaker::feasible_break_before_this_obj()` modified to move the feasible system break after the key and/or time signature when they come after a barline.

- Method `SystemLayouter::engrave_prolog()` has been modified so that the prolog can now have time signature. `StaffObjsCursor` modified to provide the applicable time signature for the prolog.

- Method `SpAlgGourlay::determine_penalty_for_line()` has been modified to remove the penalty for a system break placed after a key or time signature after a barline.

- A test score [50400-time-key-after-break.xml.txt](https://github.com/lenmus/lomse/files/7230021/50400-time-key-after-break.xml.txt) with several time/key signature changes has been added to the test-scores folder, for visual regression tests.

- ImoBarline has now a flag to mark if a key and/or time signature is defined after it. This flag is instantiated in ColStaffObjs builder. But, finally, this flag was not needed in this patch. I've preferred to keep this code and its unit test because it is quite probably that they will be needed when `<print new-system="yes">`elements are imported. I will remove this code when `<print new-system="yes">`elements are imported, if confirmed that they are finally not needed.

**Other changes, non-related to the fixed issue**

- Now, the key signature is not added to the prolog when it is an empty shape (C major or A minor keys, num. fifths == 0), because it increased blank space after the clef and there were too much space.

@dmitrio95 The issue you reported here https://github.com/lenmus/lomse/pull/320#issuecomment-923460683 should now be fixed.


dmitrio95:
Yes, this PR indeed fixes the issue I have reported but it seems to cause a new issue to appear. In the score where I have noticed this issue there is a "TRIO" text which marks the beginning of the trio part of the score (I have removed the redundant tempo marking seen in my comment in https://github.com/lenmus/lomse/pull/320#issuecomment-923460683):
изображение

This PR fixes the key signature issue but the "trio" text is now displayed at a completely unexpected position:
изображение

Perhaps the changes in the line breaking algorithm somehow cause the relevant direction object to move to a wrong system? That doesn't explain its Y position though.

The score I have used for the screenshots: [Maple_Leaf_Rag.musicxml.txt](https://github.com/lenmus/lomse/files/7231251/Maple_Leaf_Rag.musicxml.txt)

cecilios:
Thank you! And thanks for the score. I've found the problem: Other non-timed objects (e.g. the TRIO `<direction>`) defined after the key/time signature are also included before the system break. Thus the TRIO word is rendered in previous system. I do not understand why it is placed under the staves instead of above them, but this is not relevant here.

Therefore, this PR is not yet mergeable. I will work on the spacing algorithm to fix this issue.

cecilios:
**Additional Changes**
The spacing algorithm has been modified. Now key & time signatures defined just after a barline are added to the TimeSliceBarline instead to a new TimeSliceNonTimed. This ensures a clean cut when the lines break algorithm decides to place a system break after the barline and makes a clear separation of barlines (with any cautionary key/time additional shape) from other objects defined after the Key and Time signatures.

If no new issues, this can be merged.

dmitrio95:
Thanks, I see no new issues with this version of the PR, and I can confirm the issue I have mentioned is fixed now.

cecilios:
Thank you. Then, I will merge this.

PR #331, opened 2021-11-14
Fix a crash on fixing a chord-rest collision due to a missing pointer to the chord's base note

dmitrio95:
This PR fixes a crash which happens on engraving this test score: [noterest-collision-fixer-crash.musicxml.txt](https://github.com/lenmus/lomse/files/7534155/noterest-collision-fixer-crash.musicxml.txt)
 
The crash happens at [this line](https://github.com/lenmus/lomse/blob/ca924db18798258d1d3b1e61cb2a4c23c1997442/src/graphic_model/layouters/lomse_noterests_collisions_fixer.cpp#L295) because the collision fixer tries to access the pointer to the chord's base note which is null in this case. This happens because this pointer is set only for some notes in a chord (which are marked as "start", "link" or "flag" notes) and for some notes this pointer may be left unassigned. The proposed fix is to explicitly set the pointer to the base note for all notes in a chord so that the collision fixer or any other parts of the engraving algorithm can always access the base note by that pointer.

To avoid duplication I have also removed `set_base_note_shape()` calls when setting start, link and flag notes for a base note but this is not a necessary part of this patch.

### DIFF
--- a/include/lomse_shape_note.h
+++ b/include/lomse_shape_note.h
@@ -151,6 +151,7 @@ public:
 
     //for chords
     inline GmoShapeChordBaseNote* get_base_note_shape() { return m_pBaseNoteShape; }
+    inline void set_base_note_shape(GmoShapeChordBaseNote* pShape) { m_pBaseNoteShape = pShape; }
 
     //used for debug
     void set_notehead_color(Color color);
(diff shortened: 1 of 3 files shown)


cecilios:
Perhaps you already have deduced the purpose of identifying and marking "start", "link" and "flag" notes. If not, I attach some information about the chord shape.

[Lomse_Hacking_Guide--Graphic_model--Chords.pdf](https://github.com/lenmus/lomse/files/7534225/Lomse_Hacking_Guide--Graphic_model--Chords.pdf)

cecilios:
The changes look correct to me and all test passed.

I agree with the changes. Before adding the 'collision fixer' the pointer was not needed in other notes. But clearly, it is better to explicitly set the pointer to the base note for all notes in a chord to avoid future problems with new code.

Merging! Thank you very much for the PR.

PR #336, opened 2021-12-03
Add support for fingering and import it from MusicXML

cecilios:
This PR adds support for importing MusicXML fingering information and displaying it.

From the MusicXML spec it is not clear what is the expected valid content for the `<fingering>` element and what to do when a note has many `<fingering>`elements. Also, I do not have tests scores other than those created by myself. Thus, support is currently restricted to ‘expected’ and ‘logical’ combinations. Apart from numbers 1 to 5, also the fingering symbols included in SMuFL are supported (digits 0, 6 to 9, and some characters). Score [50060-fingering.xml.txt](https://github.com/lenmus/lomse/files/7650760/50060-fingering.xml.txt) is an example of currently supported fingerings:

image

**Other changes**

- `Bravura.otf` font file updated to latest version (1.392). Previous Bravura version (1.18) included with Lomse did not have glyphs for fingering.
- Fixed bug: when a glyph did not exist in the font the score was displayed as a blank page. Now the score is rendered but the missing glyph is not displayed.
- Class name `ImoAttr` renamed as `AttrObj`, as it is not an `Imo` but an auxiliary class
- Destructor of `AttrObj` now deletes the list of attributes. `ImoObj` no longer does it.
- Added regression test score 50060-fingering.xml

**IMPORTANT**

- It is necessary to install the updated Bravura font. Otherwise, fingering will not be displayed. Apart from the `Bravura.otf` file included with Lomse, there is a `.deb` installer [here](https://github.com/lenmus/bravura/releases).


dmitrio95:
If that would be of help, I have recently seen [this score](https://musescore.com/user/75786/scores/1518501) which has quite a lot of fingerings. However there doesn't seem to be anything unusual there: `<fingering>` tags contain just numbers as their content and sometimes `placement` as their attributes. I can search for more examples though if needed.

cecilios:
Thank you. I have downloaded it and tested and it is rendered fine. This PR deals with all normal cases for piano and my doubts are about other use cases.

The issue with `<fingering>` is that MusicXML documentation is not good and, thus, it is uncertain what to expect. I have open an issue ( https://github.com/w3c/musicxml/issues/437#issuecomment-985780017) and will wait for confirmation before merging this PR.

cecilios:
It seems that the clarifications from MusicXML developers will take some time.  Thus, I'm going to merge this.

PR #346, opened 2022-03-19
Support to render tablature notation

### LINKED ISSUE #344: Support tablature notation
staves score parsing and rendering no problem.
 but how to guitar score parsing rendering
gutascore
statvasscore

cecilios:
Closes #344. With this PR Lomse now properly renders tablature notation. For example:

image

But this PR does no't add support all notation related to tablature and guitar. MusicXML supports a lot of technical notations for many instruments, but as many notation applications, Lomse only renders the most common ones. Please, fell free to request support for a specific technical mark by opening an issue.

I do not have knowledge about tablature notation for guitar and, thus, I do not know which are the most common tecnical notations for tablature that should be supported.


dmitrio95:
It might probably be better to make a visual gap in staff lines around numbers written on the tablature instead of writing numbers just on top of staff lines. Most tablature editors do that by default, and that seems to look nicer because it avoids collisions of numbers with staff lines:
https://upload.wikimedia.org/wikipedia/commons/9/93/Guitar_tablature_sample.svg

If the painting order of objects is maintained, this effect can probably be achieved simply by drawing a white rectangle at the note's position before drawing the number glyph itself.

cecilios:
@dmitrio95  Thanks Dmitri. It is a good idea. It is necessary to derive from `GmoShapeNotehead` and override the `on_draw()` method to a white rectangle before drawing the glyph. I will do it now.

cecilios:
Much better now:

image

